PR #8, opened 2022-01-31
Rasa3

nelealbers:
#173. 

Updated to Rasa 3.0.5. Changes are primarily the config.yml and the way slots in domain.yml are defined. To make the installation possible, I also had to downgrade pip and sanic in the Dockerfile for Rasa_Bot.


Rasa_Bot/Dockerfile
```diff
@@ -5,6 +5,9 @@ FROM python:3.8
 # Change to root user to install dependencies
 USER root
 
+RUN python -m pip install --upgrade pip==20.2
+RUN python -m pip install --upgrade sanic==21.9.3
+
```
svenvanderburg, line 10: Could you specify this instead in the requirements.txt?

Rasa_Bot/domain.yml
```diff
@@ -51,15 +59,20 @@ slots:
...
   success_save_calendar_plan_week:
     type: bool
     influence_conversation: true
+    mappings:
+    - type: from_text
   # Evaluate completed PA
   pa_evaluation_response:
     type: float
     influence_conversation: false
-    auto_fill: false
```
svenvanderburg, line 62: Just to double-check: Did you mean to remove this `auto_fill: false`? And will it still work the same way without?

svenvanderburg (approved):
In general this looks really good! Nice that the models now have names (definitely like brutal harp), indeed it was quite difficult to distinguish based on the long timestamp strings. I left a few minor comments that would be good to check, but after that please merge this into main!

Rasa_Bot/domain.yml
nelealbers: @svenvanderburg This flag no longer exists.

PR #9, opened 2022-01-31
Less epochs for training TED

svenvanderburg:
TED policy training already converges to near 100% accuracy after 30/40 epochs. With 50 epochs we are save. This saves a lot of time when training rasa core.


nelealbers:
@svenvanderburg Looks good, and good insight!

nelealbers (approved):
@svenvanderburg Nice observation! Looks good.

PR #10, opened 2022-02-01
Spacy

nelealbers:
This adds a spacy language model to the pipeline in Rasa. Also removed entity extraction from the pipeline as we do not use that at the moment.

Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/150

### DIFF
--- a/Rasa_Bot/README.md
+++ b/Rasa_Bot/README.md
@@ -69,7 +69,10 @@ The timeout is currently set to 5 minutes (in the "domain.yml"-file). This is th
 In case the new custom action code requires any libraries, these need to be added to "requirements-actions.txt" in the "actions"-folder.
 
 ### Spacy language embeddings 
-We now use spacy language embeddings (specified in config.yml). To train the model, you now need to have spacy installed as well as the Dutch language model that we use. Have a look [here](https://spacy.io/usage) for how to install spacy.
+We now use spacy language embeddings (specified in config.yml). This means that for training a model, you now need to download the spacy language model that we use:
+
+`python -m spacy download nl_core_news_sm`
+`python -m spacy link nl_core_news_sm nl`
 
 ### Retraining when making changes to Language Model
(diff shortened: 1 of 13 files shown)


Rasa_Bot/config.yml
```diff
@@ -18,7 +24,8 @@ pipeline:
     max_ngram: 4
   - name: DIETClassifier
     epochs: 100
-  - name: EntitySynonymMapper
+    entity_recognition: False  # We do not currently have entities
```
svenvanderburg, line 27: Nice that we can turn this off :)

Rasa_Bot/README.md
```diff
@@ -68,5 +68,8 @@ The timeout is currently set to 5 minutes (in the "domain.yml"-file). This is th
 ### Add dependencies for action server to requirements
 In case the new custom action code requires any libraries, these need to be added to "requirements-actions.txt" in the "actions"-folder.
 
+### Spacy language embeddings 
+We now use spacy language embeddings (specified in config.yml). To train the model, you now need to have spacy installed as well as the Dutch language model that we use. Have a look [here](https://spacy.io/usage) for how to install spacy.
```
svenvanderburg, line 72: I am thinking maybe we can add more specific instructions here? Spacy will be installed with the `requirements.txt`. The only remaining thing is:
```
python -m spacy download nl_core_news_sm
python -m spacy link nl_core_news_sm nl
```
right? 

And that is actually not immediately clear from the link to spacy website.

svenvanderburg (approved):
Nice. I am happy you looked into this, seems like a nice update of the rasa pipeline. This would also potentially be interesting to demo to the consortium @bscheltinga, you can present it as: 'we improved the way rasa understands human language, it now uses knowledge that is has learned from a huge amount of dutch texts (for example from wikipedia)'.

Some comments:
* You uploaded 3 models, I guess we only need one
* Are you sure the uploaded models work with rasa 3.0?
* You deleted `rasa.db`. Was that intentional?
* Talk to @bscheltinga before merging, you are both pushing a new model to the main branch.

PR #11, opened 2022-02-01
Remove sender id slot

bscheltinga:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/155


Rasa_Bot/tests/test_conversations.yml
```diff
@@ -1,4 +1,4 @@
-version: "2.0"
+version: "3.0"
```
svenvanderburg, line 1: Did Nele forget this in updating to 3.0? Nice catch :)

svenvanderburg (approved):
Nice! This simplifies things a lot!

To double check: is the model you included a rasa 3.0 model?

Rasa_Bot/tests/test_conversations.yml
bscheltinga: Yes, it gave an error with the "Rasa CI / Train and Test model" test. But I tried to revert the change as I thought it should be another issue. But apparently without succes...

bscheltinga: And I should check whether I did use rasa v3 and maybe retrain the models.

svenvanderburg: No worries, just leave the change in this PR.

bscheltinga: The thing was actually that I did not use rasa 3. I retrained the model with v3 and tested it again

PR #13, opened 2022-02-07
Yes no

nelealbers:
Fixes #https://github.com/PerfectFit-project/virtual-coach-issues/issues/177.

Just using a single form made everything very unstable: slightly different training configurations would break things.

So the minimum required for helping rasa disambiguate smoking vs. PA was to duplicate two actions (always the last one at the deny/confirm blocks). 

Please have a look at what you think: does this help? I personally do think it removes duplication a bit, albeit not completely.

### DIFF
--- a/Rasa_Bot/actions/actions.py
+++ b/Rasa_Bot/actions/actions.py
@@ -207,11 +207,20 @@ class ActionResetPickedWordsSlot(Action):
         return [SlotSet("picked_words", None)]
 
 
-class ActionResetWhyPickedWordsSlot(Action):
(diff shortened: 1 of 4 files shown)


svenvanderburg (approved):
Thanks for looking into this! I agree this is a little more modular, so nice addition :)

Two minor comments:
* Did you mean to commit the `rasa.db` file?
* The linter is failing (see actions), please fix that before merging! (The rasa validation action also fails but that will be fixed with https://github.com/PerfectFit-project/virtual-coach-issues/issues/191

PR #14, opened 2022-02-10
See myself

nelealbers:
Fixes #[https://github.com/PerfectFit-project/virtual-coach-issues/issues/178](url)

### DIFF
--- a/Rasa_Bot/actions/actions.py
+++ b/Rasa_Bot/actions/actions.py
@@ -270,3 +270,51 @@ class ValidateConfirmWordsForm(FormValidationAction):
             dispatcher.utter_message("Geef alsjeblieft antwoord met 'ja' of 'nee'?")
 
         return {"confirm_words_response": yes_or_no_response}
+
+
+def validate_see_myself_as_picked_words_smoker(value):
+    if value in ["1", "2", "3", "(1)", "(2)", "(3)", 
+                 "een roker", "roker", "een niet-roker", 
+                 "iemand die stopt met roken"]:
+        return True
+    return None
+    
+
+class ValidateSeeMyselfAsSmokerForm(FormValidationAction):
+    def name(self) -> Text:
+        return 'validate_see_myself_as_smoker_form'
+
+    def validate_see_myself_as_smoker_response(
+            self, value: Text, dispatcher: CollectingDispatcher,
+            tracker: Tracker, domain: Dict[Text, Any]) -> Dict[Text, Any]:
+        """Validate picked_words_smoker input."""
+
+        picked_words_smoker = validate_see_myself_as_picked_words_smoker(value)
+        if picked_words_smoker is None:
+            dispatcher.utter_message("Antwoord alsjeblieft met 1, 2, of 3.")
+
+        return {"see_myself_as_picked_words_smoker": picked_words_smoker}
+
+
+def validate_see_myself_as_picked_words_mover(value):
+    if value in ["1", "2", "3", "(1)", "(2)", "(3)", 
+                 "lichamelijk actief", "een beetje lichamelijk actief",
+                 "niet lichamelijk actief"]:
+        return True
+    return None
+
+
+class ValidateSeeMyselfAsMoverForm(FormValidationAction):
+    def name(self) -> Text:
+        return 'validate_see_myself_as_mover_form'
+
+    def validate_see_myself_as_mover_response(
+            self, value: Text, dispatcher: CollectingDispatcher,
+            tracker: Tracker, domain: Dict[Text, Any]) -> Dict[Text, Any]:
+        """Validate picked_words_mover input."""
+
+        picked_words_mover = validate_see_myself_as_picked_words_mover(value)
+        if picked_words_mover is None:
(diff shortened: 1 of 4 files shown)


nelealbers:
@svenvanderburg  This was quite tedious. An important insight is that the slot_was_set events should be added to the stories also when setting slots to null.

Rasa_Bot/actions/actions.py
```diff
@@ -270,3 +270,118 @@ def validate_confirm_words_response(
...
+        try:
+            value = int(value)
+        except ValueError:
+            return False
+        if (value < 1) or (value > 3):
+            return False
+        return True
+
+
+class ActionMapSeeMyselfAsPickedWordsSmoker(Action):
```
svenvanderburg, line 302: Do you have a good reason for using a separate mapping action. We could also set 2 slots in the `ValidateSeeMyselfAsSmokerForm`? I am not sure if that is nicer. Or use a text slot instead of an integer slot for storing the answer. Or maybe we should just learn to live with needing this kind of very verbose code with rasa...

Rasa_Bot/config.yml
```diff
@@ -40,7 +40,7 @@ policies:
     max_history: 7
   - name: TEDPolicy
     max_history: 14
-    epochs: 50
+    epochs: 150
```
svenvanderburg, line 43: I guess TED needs more epochs now to converge?

Rasa_Bot/data/stories.yml
```diff
@@ -132,7 +132,15 @@ stories:
     - active_loop: null  # TODO: store slot in database
     - action: utter_why_picked_words_response
     - action: action_reset_picked_words_slot
+    - slot_was_set:
+      - picked_words: null
```
svenvanderburg, line 136: Wow, it is really cumbersome that we have to add this... Thanks for investigating!

Rasa_Bot/domain.yml
```diff
@@ -212,9 +224,10 @@ responses:
              Welke woorden passen bij jou als roker?
              Roken is...
             "
+  utter_intro_current_mover:
+    - text: "We gaan nu dezelfde oefening doen voor bewegen."
   utter_current_mover:
-    - text: "We gaan nu dezelfde oefening doen voor bewegen.
-             Welke woorden vind jij passen bij bewegen?
+    - text: "Welke woorden vind jij passen bij bewegen?
```
svenvanderburg, line 230: Nice improvement!

Rasa_Bot/domain.yml
```diff
@@ -251,28 +264,37 @@ responses:
...
+  utter_see_myself_as_picked_smoker_words:
+    - text: "{see_myself_as_picked_words_smoker} heb jij gekozen, klopt dat?"
+  utter_see_myself_as_picked_mover_words:
+    - text: "{see_myself_as_picked_words_mover} heb jij gekozen, klopt dat?"
   utter_thank_user_for_response:
     - text: "Helder, bedankt."
+  utter_leuk:
+    - text: "Leuk!"
+  utter_fijn:
+    - text: "Fijn!"
```
svenvanderburg, line 285: Can we maybe make this into a single utterance `utter_express_fun` (There's probably a better word), with multiple `texts` 'Leuk!', 'Fijn!' and maybe more?

Rasa_Bot/domain.yml
```diff
@@ -251,28 +264,37 @@ responses:
...
   utter_ask_why_see_myself_as_smoker_form_why_picked_words:
-    - text: "Leg uit, waarom zie jij jezelf als {picked_words_smoker}?" # TODO: use generic slot?
+    - text: "Leg uit, waarom zie jij jezelf als {see_myself_as_picked_words_smoker_text}?" # TODO: use generic slot?
   utter_ask_why_see_myself_as_mover_form_why_picked_words:
-    - text: "En waarom zie jij jezelf als {picked_words_mover}?" # TODO: which slot to use here?
+    - text: "En waarom zie jij jezelf als {see_myself_as_picked_words_mover_text}?" # TODO: which slot to use here?
```
svenvanderburg, line 289: I think the comment can be removed? Because we use this slot now?

Rasa_Bot/domain.yml
```diff
@@ -251,28 +264,37 @@ responses:
...
   utter_close_future_self_step_1:
     - text: "Heldere antwoorden. Dankje voor het delen.
              We gaan door met nieuwe oefeningen."
+  utter_try_again:
+    - text: "Oke laten we het opnieuw proberen."
+  utter_try_again_see_myself_mover:
+    - text: "Oke laten we het nog een keer proberen."
```
svenvanderburg, line 296: I have the feeling we need these 2 utterances with different names because otherwise rasa gets confused of where we are in the conversation? Otherwise I propose on utterance with 2 texts.

svenvanderburg (approved):
Looks good @nelealbers ! Indeed looks rather tedious. This dialog is becoming quite a monster... But I think this is the right way to implement it. I left a few comments, please merge if you have incorporated them!

Rasa_Bot/domain.yml
nelealbers: Yes good point

PR #17, opened 2022-02-16
Ask for more words if not enough, (with action-test) in future self dialog

drcandacemakedamoore:
Merge from Makeda and Nele branches, action test- rewritten.

### DIFF
--- a/Rasa_Bot/tests/test_actions.py
+++ b/Rasa_Bot/tests/test_actions.py
@@ -68,7 +68,7 @@ has_enough_words_examples = (
     has_enough_words_examples,
     ids=("negative", "positive"),
 )
-async def test_run_action_validate_has_enough_words_form(
+async def test_run_action_validate_why_picked_mover_words_form(
         dispatcher: CollectingDispatcher,
         domain: DomainDict,
         response: Text,
--- a/Rasa_Bot/tests/test_actions.py
+++ b/Rasa_Bot/tests/test_actions.py
@@ -1,6 +1,8 @@
 """Unit tests for the custom actions"""
 import pytest
 from unittest import mock
+from contextlib import contextmanager
+from typing import Text, Union
 from rasa_sdk.events import SlotSet
 from rasa_sdk.executor import CollectingDispatcher
 from rasa_sdk.types import DomainDict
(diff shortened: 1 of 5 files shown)


Rasa_Bot/tests/test_actions.py
```diff
@@ -40,3 +42,46 @@ async def test_run_action_store_pa_evaluation(
     expected_events = [SlotSet("pa_evaluation_response", None)]
     assert events == expected_events
     mock_session.commit.assert_called_once()
+
+
+# TODO: If this is a recurring pattern, this can be turned into a fixture
```
nelealbers, line 47: What do you mean here? If it is a recurring pattern to need to create a previous message?

Rasa_Bot/tests/test_actions.py
```diff
@@ -40,3 +42,46 @@ async def test_run_action_store_pa_evaluation(
...
+@contextmanager
+def latest_message(tracker, message):
```
nelealbers, line 49: I see, so instead of adapting the tracker file itself, you just change the last message in the tracker after having imported it. I think that is a good idea, since working with the tracker file directly may be rather cumbersome.

Rasa_Bot/tests/test_actions.py
```diff
@@ -40,3 +42,46 @@ async def test_run_action_store_pa_evaluation(
...
+    ),
+)
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "response,expected",
+    has_enough_words_examples,
+    ids=("negative", "positive"),
+)
+async def test_run_action_validate_has_enough_words_form(
```
nelealbers, line 71: Would be good to adapt the naming of this function to the name of the form that you are testing, so test_run_action_validate_why_picked_mover_words_form.

nelealbers (commented):
Looks good, and nice approach with replacing the last message from the tracker. Once you incorporate my comments and fix the linter complains this should be good to go.

Rasa_Bot/tests/test_actions.py
drcandacemakedamoore: If there will be  multiple tests that use tracker in the same way, then it can become a "fixture" which is a specific entity in pytest.

drcandacemakedamoore:
I notice you guys simply merge. I'm going for a "rebase and merge" aka a rebase in order to keep the repository history cleaner.

PR #18, opened 2022-03-03
Rescheduling dialog

nelealbers:
Fixes #https://github.com/PerfectFit-project/virtual-coach-issues/issues/179

### DIFF
--- a/.prospector.yml
+++ b/.prospector.yml
@@ -14,12 +14,10 @@ ignore-paths:
 pyroma:
     run: true
 
-pep8:
+pydocstyle:
     full: true
     options:
       max-line-length: 100
-
-pep257:
     disable: [
         # Disable because not part of PEP257 official convention:
         # see http://pep257.readthedocs.io/en/latest/error_codes.html
--- a/.prospector.yml
+++ b/.prospector.yml
@@ -14,10 +14,12 @@ ignore-paths:
 pyroma:
     run: true
 
-pydocstyle:
+pycodestyle:
     full: true
     options:
       max-line-length: 100
+    
+pydocstyle:
     disable: [
         # Disable because not part of PEP257 official convention:
         # see http://pep257.readthedocs.io/en/latest/error_codes.html
--- a/Rasa_Bot/actions/actions.py
+++ b/Rasa_Bot/actions/actions.py
@@ -272,7 +272,7 @@ class ValidateConfirmWordsForm(FormValidationAction):
 
         return {"confirm_words_response": yes_or_no_response}
 
-
+    
 class ActionResetReschedulingNowSlot(Action):
     """Reset rescheduling_now slot"""
 
--- a/Rasa_Bot/actions/actions.py
+++ b/Rasa_Bot/actions/actions.py
@@ -142,7 +142,7 @@ class ValidatePaEvaluationForm(FormValidationAction):
         """Validate pa_evaluation_response input."""
 
         if not self._is_valid_input(value):
-            dispatcher.utter_message("Kun je een geheel getal tussen 1 en 5 opgeven?")
+            dispatcher.utter_message(template="utter_please_answer_1_to_5")
             return {"pa_evaluation_response": None}
         pa_evaluation_response = int(value)
         return {"pa_evaluation_response": pa_evaluation_response}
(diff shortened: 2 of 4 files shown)


Rasa_Bot/data/stories.yml
```diff
@@ -97,14 +97,62 @@ stories:
...
+    - action: rescheduling_now_or_later_form
+    - active_loop: rescheduling_now_or_later_form
+    - slot_was_set: 
+      - rescheduling_now: false
+    - active_loop: null
+    - action: utter_user_chose_later
+    - action: action_reset_rescheduling_now_slot
+    - slot_was_set:
+      - rescheduling_now: null
+    - action: rescheduling_options_form  # TODO: determine suitable rescheduling options.
```
svenvanderburg, line 131: Is this something we need to discuss during the demo?

Rasa_Bot/data/stories.yml
```diff
@@ -97,14 +97,62 @@ stories:
...
+    - active_loop: rescheduling_options_form  # TODO: use rescheduling_option slot to reschedule
```
svenvanderburg, line 132: Can you create a follow-up issue for this?

Rasa_Bot/data/stories.yml
```diff
@@ -97,14 +97,62 @@ stories:
...
+    - active_loop: null
+    - action: utter_rescheduling_option_chosen
+    - action: action_reset_rescheduling_option_slot
+    - slot_was_set:
+      - rescheduling_option: null
+    # TODO: what are we going to do at the end of the dialog if the user types something again?
+    # Maybe a ConversationPaused() event from custom action?
```
svenvanderburg, line 139: That's a good question. I guess now the coach will just try to respond to whatever the user says. Which might be OK, but can also lead to strange conversations I guess. If we do a `ConversationPaused` event we also need to resume it again at some point. I am not sure whether that will help, at least it will make things a bit more complicated.

Maybe we can demo a case where the conversation becomes a bit weird and see how bad it is and what to do about it?

Rasa_Bot/domain.yml
```diff
@@ -181,35 +200,72 @@ responses:
...
+
+  ### Re-prompts in forms
+  utter_please_answer_1_2_3:
+     - text: "Antwoord alsjeblieft met '1,' '2,' of '3.'"
+  utter_please_answer_1_to_5:
+     - text: "Kun je een geheel getal tussen 1 en 5 opgeven?"
+  utter_please_answer_yes_no:
+     - text: "Geef alsjeblieft antwoord met 'ja' of 'nee'?"
+  utter_did_not_understand:
+     - text: "Hmm ik heb dat niet begrepen."
```
svenvanderburg, line 228: Cool, I quite like this organisation. Better that this lives here and not in the code.

Rasa_Bot/actions/actions.py
```diff
@@ -142,7 +142,7 @@ def validate_pa_evaluation_response(
         """Validate pa_evaluation_response input."""
 
         if not self._is_valid_input(value):
-            dispatcher.utter_message("Kun je een geheel getal tussen 1 en 5 opgeven?")
+            dispatcher.utter_message(template="utter_please_answer_1_to_5")
```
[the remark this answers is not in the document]
svenvanderburg: OK, I see this is now in the `domain.yml`. I quite like that! 👍

Rasa_Bot/actions/actions.py
```diff
@@ -268,11 +268,83 @@ def validate_confirm_words_response(
...
+
+    def validate_rescheduling_now(
+            self, value: Text, dispatcher: CollectingDispatcher,
+            tracker: Tracker, domain: Dict[Text, Any]) -> Dict[Text, Any]:
+        # pylint: disable=unused-argument
+        """Validate rescheduling_now input."""
+
+        now_or_later = validate_now_or_later_response(value)
+        if now_or_later is None:
+            dispatcher.utter_message("Geef alsjeblieft antwoord met 'nu' of 'later.'")
```
svenvanderburg, line 306: Maybe here we can also make use of the `template`?

Rasa_Bot/actions/actions.py
```diff
@@ -268,11 +268,83 @@ def validate_confirm_words_response(
...
+    """Reset rescheduling_now slot"""
+
+    def name(self):
+        return "action_reset_rescheduling_now_slot"
+
+    async def run(self, dispatcher, tracker, domain):
+        return [SlotSet("rescheduling_now", None)]
+    
+
+def validate_now_or_later_response(value):
```
svenvanderburg, line 286: Is this reused in other classes? Otherwise it's maybe nicer to have it as a private method in the class?

svenvanderburg (approved):
Looks quite good @nelealbers ! I left some comments, but I am pretty sure that you can figure it out. Let me know which things we should discuss during the demo!

Feel free to merge once you looked at the comments.

.prospector.yml
```diff
@@ -14,12 +14,10 @@ ignore-paths:
 pyroma:
     run: true
 
-pep8:
+pydocstyle:
     full: true
     options:
       max-line-length: 100
-
-pep257:
```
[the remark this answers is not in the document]
nelealbers: @svenvanderburg Apparently there was a prospector update so that our former prospector configuration made prospector fail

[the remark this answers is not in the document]
svenvanderburg: Cool, thanks for updating!

Rasa_Bot/data/stories.yml
nelealbers: Yes, I think so. Suitable options depend on when we start the future self dialog. Question is if that is always the same time in the day, or if it differs per person.

PR #19, opened 2022-03-03
Temporary change to rasa bot dockerfile to make it compatible with apple M1 chip

svenvanderburg:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/208

It is a bit of an ugly workaround. And we should revert this once tensorflow and rasa have fixed this in a proper way, I created https://github.com/PerfectFit-project/virtual-coach-issues/issues/209 for that.

The most important point for reviewer is to see if this still works on their machine :)


wbaccinelli:
Tested on Mac and Windows machine and working.

PR #27, opened 2022-03-15
Save future self dialog state rasa

nelealbers:
For https://github.com/PerfectFit-project/virtual-coach-issues/issues/182. 

Also fixed some things on the way, e.g., do not consider capitalization when typing "ja" or "nu", because that is annoying when using the niceday app.


svenvanderburg (approved):
Great work 👍  Nothing to add actually :) Thanks for improving, you are following the [boy scout rule](https://www.stepsize.com/blog/how-to-be-an-effective-boy-girl-scout-engineer#:~:text=Most%20engineers%20have%20heard%20of,wouldn't%20deteriorate%20so%20relentlessly.) very well :)

PR #28, opened 2022-03-18
Store future self dialog

bscheltinga:
fixes #180


Rasa_Bot/actions/actions.py
```diff
@@ -28,6 +27,25 @@
...
 
 
+class DialogQuestions(Enum):
+    FUTURE_SELF_SMOKER_WORDS = 1
+    FUTURE_SELF_SMOKER_WHY = 2
+    FUTURE_SELF_MOVER_WORDS = 3
+    FUTURE_SELF_MOVER_WHY = 4
+
+
+def store_dialog_answer_to_db(user_id, answer, question: DialogQuestions):
```
svenvanderburg, line 37: I really like the interface of this function!

Rasa_Bot/actions/actions.py
```diff
@@ -28,6 +27,25 @@
 NICEDAY_API_ENDPOINT = os.getenv('NICEDAY_API_ENDPOINT')
 
 
+class DialogQuestions(Enum):
+    FUTURE_SELF_SMOKER_WORDS = 1
+    FUTURE_SELF_SMOKER_WHY = 2
+    FUTURE_SELF_MOVER_WORDS = 3
+    FUTURE_SELF_MOVER_WHY = 4
```
svenvanderburg, line 34: Maybe you can add a tiny bit of documentation here so it is easier to remember what questions these refer to. Could be as simple as adding a comment:
`# Why did you pick these words for moving?`

Rasa_Bot/domain.yml
```diff
@@ -391,7 +391,7 @@ actions:
 - validate_pa_evaluation_form
 - action_utter_pa_evaluation_form_filled
 - action_store_pa_evaluation
-### Future self dialog
+- action_store_smoker_words
```
svenvanderburg, line 394: Would be nice to keep the comment that these actions are for future self dialog

svenvanderburg (commented):
Nice work @bscheltinga ! ⭐  See my comments in the code. Also:
* Please also store the other 3 answers
* The linter is failing, there is one line that is too long: 
image
* You committed 2 new model files, please keep only the most recent one and remove the current model file.

If you want I can take another look before you merge, but feel free to merge if you feel confident.

Rasa_Bot/domain.yml
bscheltinga: Yes, good idea

PR #31, opened 2022-03-31
Bump rasa version to 3.1.0

svenvanderburg:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/191

PR #32, opened 2022-04-01
Resched options

nelealbers:
Fixes the first element of https://github.com/PerfectFit-project/virtual-coach-issues/issues/224, i.e., rescheduling options depend on the current time and there is a 4th rescheduling option for 1h from now.

### DIFF
--- a/Rasa_Bot/actions/actions.py
+++ b/Rasa_Bot/actions/actions.py
@@ -443,17 +443,17 @@ class ActionGetReschedulingOptionsList(Action):
         current_hour = datetime.datetime.now(pytz.timezone('Europe/Amsterdam')).hour
         
         # In the morning
-        if (current_hour >= morning[0]) and (current_hour < morning[1]):
+        if morning[0] <= current_hour < morning[1]:
             options +=  ["Vanmiddag, om 16:00",
                          "Vanavond, om 21:00",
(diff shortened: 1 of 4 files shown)


Rasa_Bot/actions/actions.py
```diff
@@ -422,6 +423,57 @@ def _validate_now_or_later_response(value):
...
+    
+    def name(self):
+        return "action_get_rescheduling_options_list"
+    
+    async def run(self, dispatcher, tracker, domain):
+        
+        # define morning, afternoon, evening
+        morning = [6, 12]
+        afternoon = [12, 18]
+        evening = [18, 24]
```
svenvanderburg, line 439:
```suggestion
        # define morning, afternoon, evening
        MORNING = (6, 12)
        AFTERNOON = (12, 18)
        EVENING = (18, 24)
```
Would be nice to mark these as constants (by capitalizing), and use tuples instead of lists, because they should be unmutable.

svenvanderburg (approved):
Cool! Nice work :) I left 1 comment in the code. For the rest: Please also update the end-to-end tests in `virtual-coach-main` accordingly (not sure how to deal with the fact that the response differs depending on the time of day, probably @wbaccinelli has some pointers for that). Also I just merged #31, so you need to retrain the model before merging.

PR #34, opened 2022-04-05
Issue236 use database url

raar1:
As part of issue https://github.com/PerfectFit-project/virtual-coach-issues/issues/236 , this makes the Rasa actions scripts use the `DATABASE_URL` env instead of DB_HOST.

### DIFF
--- a/Rasa_Bot/README.md
+++ b/Rasa_Bot/README.md
@@ -10,7 +10,7 @@ See here for the instructions from Rasa: https://rasa.com/docs/rasa/docker/deplo
 The steps are as follows:
 - Make sure that you have Docker and Docker Compose installed. You can check whether you do by typing `docker -v && docker-compose -v`.
 - Start the database via docker-compose, apply the existing migrations and load test data (see [here](https://github.com/PerfectFit-project/virtual-coach-db).
-- Create a .env-file in the Rasa_Bot/actions-folder that contains the host address and port of the running database in the variable DB_HOST. If the database is running on localhost on Windows or Mac and you want to access it from inside a Docker container, set DB_Host to host.docker.internal:<port_number> (see .env-example).
+- Create a .env-file in the Rasa_Bot/actions-folder that contains the host address and port of the running database in the variable DATABASE_URL. If the database is running on localhost on Windows or Mac and you want to access it from inside a Docker container, set DATABASE_URL to host.docker.internal:<port_number> (see .env-example).
 - Navigate to the "Rasa_Bot"-folder on your laptop.
 - Type `docker-compose up`.
 - Now you can communicate with the bot via its REST API. E.g. on Windows, type `curl http://localhost:5005/webhooks/rest/webhook -d "{\"message\": \"Kan ik de agenda voor de week krijgen?\", \"sender\":\"user\"}"`. Note that the escaping of the double-quotes is a fix that is needed on Windows.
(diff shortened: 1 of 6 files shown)


Rasa_Bot/README.md
```diff
@@ -10,7 +10,7 @@ See here for the instructions from Rasa: https://rasa.com/docs/rasa/docker/deplo
 The steps are as follows:
 - Make sure that you have Docker and Docker Compose installed. You can check whether you do by typing `docker -v && docker-compose -v`.
 - Start the database via docker-compose, apply the existing migrations and load test data (see [here](https://github.com/PerfectFit-project/virtual-coach-db).
-- Create a .env-file in the Rasa_Bot/actions-folder that contains the host address and port of the running database in the variable DB_HOST. If the database is running on localhost on Windows or Mac and you want to access it from inside a Docker container, set DB_Host to host.docker.internal:<port_number> (see .env-example).
+- Create a .env-file in the Rasa_Bot/actions-folder that contains the host address and port of the running database in the variable DATABASE_URL. If the database is running on localhost on Windows or Mac and you want to access it from inside a Docker container, set DATABASE_URL to host.docker.internal:<port_number> (see .env-example).
```
svenvanderburg, line 13: I am not sure how up-to-date this documentation is, but please give the full db url as an example here.

Rasa_Bot/actions/.env-example
```diff
@@ -1,2 +1,2 @@
-DB_HOST = host.docker.internal:5432
-NICEDAY_API_ENDPOINT = 'http://niceday_api:8080/'
\ No newline at end of file
+DATABASE_URL = host.docker.internal:5432/perfectfit
```
svenvanderburg, line 1: This needs to be in the full url format right?, i.e. 'postgres:username:password//'-something.

Rasa_Bot/actions/actions.py
```diff
@@ -23,9 +22,8 @@
                                               UserInterventionState)
 from virtual_coach_db.helper.helper import get_db_session
 
-# load .env-file and get db_host and niceday_api_endopint variables
-load_dotenv()
-DB_HOST = os.getenv('DB_HOST')
+# load database url and niceday_api_endopint variables
+DATABASE_URL = os.getenv('DATABASE_URL')
 NICEDAY_API_ENDPOINT = os.getenv('NICEDAY_API_ENDPOINT')
```
svenvanderburg, line 27: Why did you remove dotenv? I think it is still nice for devleopers to be able to configure these variables in a .env file?

scheduler/celery_tasks.py
```diff
@@ -2,13 +2,9 @@
 
 import requests
 from celery import Celery
-from dotenv import load_dotenv
 from virtual_coach_db.dbschema.models import Users
 from virtual_coach_db.helper.helper import get_db_session
 
-load_dotenv()
-DB_HOST = os.getenv('DB_HOST')
-
```
svenvanderburg, line 11: Also here, dotenv might still be nice!

svenvanderburg (approved):
Left some minor comments, but for the rest looks good!

Rasa_Bot/README.md
raar1: Agreed, I've updated this now to be the full url (including password, user etc)

Rasa_Bot/actions/.env-example
raar1: Updated this one and also the .env-example in the root directory (should we have so many now...?)

Rasa_Bot/actions/actions.py
raar1: I'm worried that with so many .env files lying around that at some point people will be confused as to which/where the env variables are actually coming from. For example, if in the docker-compose people set a particular set of env vars and then the script later silently loads the .env from the directory, that could lead to clashes or unexpected behaviour.

raar1: If we don't have the load dotenv in the scripts, then it ensures that people explicitly load a .env. If you use the docker-compose run method, it already does this for you, so anything is .env is already being included in the os.environment

scheduler/celery_tasks.py
raar1: Agreed but it is still there, just loaded in the docker-compose.yml

PR #35, opened 2022-04-05
Store i see myself as answers in db

bscheltinga:
fixes #220


svenvanderburg (approved):
Nice! Only 1 comment: I guess you still need to commit a newly trained model right? Feel free to merge after testing that.

PR #40, opened 2022-05-12
Future self rep

nelealbers:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/181.

Uses conditional response variations.

### DIFF
--- a/Rasa_Bot/actions/actions.py
+++ b/Rasa_Bot/actions/actions.py
@@ -732,8 +732,8 @@ class ActionSetFutureSelfDialogStateStep1(Action):
 
     async def run(self, dispatcher, tracker, domain):
         return [SlotSet("future_self_dialog_state", 1)]
-    
-    
+
+
 class ActionGetFutureSelfRepetitionFromDatabase(Action):
     """"To get from database whether this is a repetition of the
         future self dialog and if yes, the relevant saved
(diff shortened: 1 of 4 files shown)


Rasa_Bot/actions/actions.py
```diff
@@ -733,6 +734,93 @@ async def run(self, dispatcher, tracker, domain):
...
+            query_mover =  (
+                session.query(
+                   DialogAnswers
+                )
+                .filter(
+                    DialogAnswers.users_nicedayuid==user_id,
+                    DialogAnswers.question_id==DialogQuestions.FUTURE_SELF_MOVER_WORDS.value,
+                    DialogAnswers.datetime == subquery_mover
+                )
+                .first()
```
svenvanderburg, line 810: The logic for smoker & mover is exactly the same right? You could refactor it into a single function with an argument 'chosen_words_type' that toggles the functionality.

Rasa_Bot/domain.yml
```diff
@@ -318,15 +389,61 @@ responses:
...
+          value: true
+      text: "Eerst kijken wij wat voor roker jij op dit moment bent.\n
+             De vorige keer zei je dit over roken: {future_self_dialog_smoker_words_prev}.
+             Hoe kijk je er nu tegenaan?\n
+             Roken is ...
+            "
+    - condition: 
+        - type: slot
+          name: future_self_dialog_step_1_repetition
+          value: true
```
svenvanderburg, line 413: Nice that this works well like this!

svenvanderburg (commented):
Nice work @nelealbers ! I left a few comments. Did you also already made changes in the end2end tests? It would be nice to add a test that tests the repetition case.

Rasa_Bot/actions/actions.py
svenvanderburg: Also, I think it would be wise to add a unit test for this query.

PR #42, opened 2022-06-06
Get first aid kit

nelealbers:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/254 (tasks 2 and 3).

### DIFF
--- a/Rasa_Bot/requirements.txt
+++ b/Rasa_Bot/requirements.txt
@@ -1,5 +1,5 @@
 git+https://github.com/PerfectFit-project/niceday_client
-protobuf==3.20.1
+protobuf==3.20.1  # temporary fix
 rasa==3.1.0 # NB! when updating, make sure to also update:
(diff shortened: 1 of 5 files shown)


Rasa_Bot/requirements.txt
```diff
@@ -1,5 +1,7 @@
 git+https://github.com/PerfectFit-project/niceday_client
+protobuf==3.20.1  # temporary fix
```
wbaccinelli, line 2: Is this package something we will not use? In this case just remember to open an issue for fixing it

wbaccinelli (approved):
Very nice!

Rasa_Bot/requirements.txt
nelealbers: @wbaccinelli This is a temporary fix because I could not longer build the rasa server Docker image. So good point to create an issue for this!

PR #47, opened 2022-06-23
249 create preparation phase dialogs trigger

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/249


wbaccinelli:
checks should be fixed when [this](https://github.com/PerfectFit-project/virtual-coach-db/pull/29) pull is merged in the main branch

scheduler/celery_tasks.py
```diff
@@ -17,6 +29,38 @@
     },
 }
 
+@app.task
+def dialog_completed(user_id: int, dialog_name: str):
+    phase = get_current_phase(user_id)
+    dialog_id = get_intervention_component_id(dialog_name)
```
svenvanderburg, line 35: I think we should settle on calling stuff either `intervention_component` or `dialog` but now we use them interchangably and that is confusing.

Rasa_Bot/actions/actions.py
```diff
@@ -906,4 +910,70 @@ async def run(self, dispatcher, tracker, domain):
...
+    async def run(self, dispatcher, tracker, domain):
+        return [SlotSet("current_intervention_component", PreparationDialogs.COLD_TURKEY)]
+
+
+class SetSlotPlanQuitStart(Action):
+    def name(self):
+        return "action_slot_plan_quit_start"
+
+    async def run(self, dispatcher, tracker, domain):
+        return [SlotSet("current_intervention_component", PreparationDialogs.PLAN_QUIT_START_DATE)]
```
svenvanderburg, line 963:
```suggestion
class SetSlotPlanQuitStartDate(Action):
    def name(self):
        return "action_slot_plan_quit_start_date"

    async def run(self, dispatcher, tracker, domain):
        return [SlotSet("current_intervention_component", PreparationDialogs.PLAN_QUIT_START_DATE)]
```
Would be good to have consistent naming

Rasa_Bot/actions/actions.py
```diff
@@ -906,4 +910,70 @@ async def run(self, dispatcher, tracker, domain):
...
+
+
+class SetSlotMentalContrasting(Action):
+    def name(self):
+        return "action_slot_mental_contrasting"
+
+    async def run(self, dispatcher, tracker, domain):
+        return [SlotSet("current_intervention_component", PreparationDialogs.FUTURE_SELF)]
```
svenvanderburg, line 971:
```suggestion
class SetSlotMentalContrasting(Action):
    def name(self):
        return "action_set_slot_mental_contrasting"

    async def run(self, dispatcher, tracker, domain):
        return [SlotSet("current_intervention_component", PreparationDialogs.FUTURE_SELF)]
```

Rasa_Bot/actions/actions.py
```diff
@@ -906,4 +910,70 @@ async def run(self, dispatcher, tracker, domain):
...
+
+
+class SetSlotGoalSetting(Action):
+    def name(self):
+        return "action_slot_goal_setting"
+
+    async def run(self, dispatcher, tracker, domain):
+        return [SlotSet("current_intervention_component", PreparationDialogs.GOAL_SETTING)]
```
svenvanderburg, line 979:
```suggestion
class SetSlotGoalSetting(Action):
    def name(self):
        return "action_set_slot_goal_setting"

    async def run(self, dispatcher, tracker, domain):
        return [SlotSet("current_intervention_component", PreparationDialogs.GOAL_SETTING)]
```

svenvanderburg: I suggest to do this also for the other ones, it makes it a little bit clearer.

scheduler/celery_tasks.py
```diff
@@ -17,6 +29,38 @@
...
+
+    if phase.phase_name == Phases.PREPARATION:
+        next_dialog = get_next_preparation_dialog(dialog_name)
+
+        if next_dialog is not None:
+            endpoint = f'http://rasa_server:5005/conversations/{user_id}/trigger_intent'
+            headers = {'Content-Type': 'application/json'}
+            params = {'output_channel': 'niceday_input_channel'}
+            data = '{"name": "' + next_dialog[1] + '" }'
+            requests.post(endpoint, headers=headers, params=params, data=data)
```
svenvanderburg, line 48: You could refactor this in a function `trigger_dialog`

scheduler/celery_tasks.py
```diff
@@ -37,6 +81,107 @@ def get_user_ids():
...
+    if dialog_id == PreparationDialogs.COLD_TURKEY:
+        next_dialog = [PreparationDialogs.PLAN_QUIT_START_DATE,
+                       PreparationDialogsTriggers.PLAN_QUIT_START_DATE.value]
+    if dialog_id == PreparationDialogs.PLAN_QUIT_START_DATE:
+        next_dialog = [PreparationDialogs.FUTURE_SELF,
+                       PreparationDialogsTriggers.FUTURE_SELF.value]
+    if dialog_id == PreparationDialogs.FUTURE_SELF:
+        next_dialog = [PreparationDialogs.GOAL_SETTING,
+                       PreparationDialogsTriggers.GOAL_SETTING.value]
+    if dialog_id == PreparationDialogs.GOAL_SETTING:
```
svenvanderburg, line 130: You could also make a list of the different steps in the correct order, and then take the next item in the list.

scheduler/celery_tasks.py
wbaccinelli: That's a good point, I'm changing all to intervention_component

scheduler/celery_tasks.py
wbaccinelli: Here I'm not sure if having a task that launches a task immediately would have some drawback effects.

scheduler/definitions.py
```diff
@@ -0,0 +1,25 @@
...
+    GOAL_SETTING = 'goal_setting'
+
+
+class PreparationDialogsTriggers(str, Enum):
+    PROFILE_CREATION = 'EXTERNAL_trigger_profile_creation'
+    MEDICATION_TALK = 'EXTERNAL_trigger_medication_talk'
+    COLD_TURKEY = 'EXTERNAL_trigger_cold_turkey'
+    PLAN_QUIT_START_DATE = 'EXTERNAL_trigger_plan_quit_start'
+    FUTURE_SELF = 'EXTERNAL_trigger_mental_contrasting'
+    GOAL_SETTING = 'EXTERNAL_trigger_goal_setting'
```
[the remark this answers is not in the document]
wbaccinelli: Yes, it is already like this, but I forgot to delete this script :)

PR #51, opened 2022-07-11
268 clean start dialogs

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/268

All the dialogs now end with the `action_end_dialog` action. This action starts the default `action_restart` to reset the conversation. In the future, this action can be used to run end of dialog routines, in case of need.

### DIFF
--- a/scheduler/celery_tasks.py
+++ b/scheduler/celery_tasks.py
@@ -114,13 +114,13 @@ def get_current_phase(user_id: int) -> InterventionPhases:
         session.query(
             UserInterventionState
         )
-            .join(InterventionPhases)
-            .filter(
+        .join(InterventionPhases)
+        .filter(
             UserInterventionState.users_nicedayuid == user_id
         )
-            .order_by(UserInterventionState.id.desc())  # order by descending id
-            .limit(1)  # get only the first result
-            .all()
+        .order_by(UserInterventionState.id.desc())  # order by descending id
+        .limit(1)  # get only the first result
+        .all()
     )
 
     phase = selected[0].phase
@@ -185,10 +185,10 @@ def get_intervention_component_id(intervention_component_name: str) -> int:
         session.query(
             InterventionComponents
         )
-            .filter(
+        .filter(
             InterventionComponents.intervention_component_name == intervention_component_name
         )
-            .all()
+        .all()
     )
 
     intervention_component_id = selected[0].intervention_component_id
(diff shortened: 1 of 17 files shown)


scheduler/celery_tasks.py
```diff
@@ -160,7 +162,7 @@ def store_intervention_component_to_db(user_id: int,
     session = get_db_session(db_url=DATABASE_URL)  # Create session object to connect db
     selected = session.query(Users).filter_by(nicedayuid=user_id).one()
 
-    entry = UserInterventionState(intervention_phase_id =intervention_phase_id,
+    entry = UserInterventionState(intervention_phase_id=intervention_phase_id,
                                   intervention_component_id=intervention_component_id,
                                   completed=completed,
                                   last_time=datetime.now().astimezone(TIMEZONE),
```
raar1, line 168: Are we assuming the user never leaves CET? Would something weird happen if they went on holiday to e.g. Portugal?

raar1: Also does timezone take into account daylight savings?

scheduler/celery_tasks.py
```diff
@@ -146,11 +147,12 @@ def schedule_intervention_component_execution(user_id: int):
             it triggers the profile creation intervention component one minute after the request
         TODO: Check DB to get the preferences, schedule all intervention components accordingly
     """
-    planned_date = datetime.now() + timedelta(minutes = 1)
+    planned_date = datetime.now() + timedelta(minutes=1)
     print(planned_date)
```
raar1, line 151: Do we want to keep these debug messages? Or put them in a log maybe

raar1 (approved):
Good work! I left a couple of minor comments.

Out of curiosity, can these actions ever fail for some reason? If action_restart never gets called then will that mess anything else up?

wbaccinelli:
In theory, this action should not fail as long as the action_server is up and running and the rasa_server is connected to that. I feel confident especially because the action_restart is a Rasa default action. 
If action_restart is not called, it should not be a big issue anyway when we will have the real dialogs in place. This is a security layer, to make sure that everything is reset before starting a new dialog.

scheduler/celery_tasks.py
wbaccinelli: Yes, timezone takes into account DST. 
Unfortunately we have to assume they will not leave CET. The problem here is that to know the actual timezone of the user we should get this information from the application itself. We had this discussion with Sven and we decided that at this point we can assume that the users will be in CET during the trial.

scheduler/celery_tasks.py
wbaccinelli: No, these messages are gong to be deleted very very soon. This is just a dummy function to be substituted already in the current sprint.

PR #53, opened 2022-07-15
252 create execution phase dialogs

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/252

### DIFF
--- a/scheduler/celery_tasks.py
+++ b/scheduler/celery_tasks.py
@@ -5,7 +5,7 @@ import requests
 from celery import Celery
 from datetime import datetime, date, timedelta
 from dateutil import tz
-from virtual_coach_db.dbschema.models import (Users, UserInterventionState,
+from virtual_coach_db.dbschema.models import (Users, UserInterventionState, UserPreferences,
                                               InterventionPhases, InterventionComponents)
 from virtual_coach_db.helper.helper_functions import get_db_session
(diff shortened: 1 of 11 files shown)


scheduler/celery_tasks.py
```diff
@@ -21,41 +20,31 @@
...
-                              PreparationInterventionComponentsTriggers.FUTURE_SELF.value,
-                              PreparationInterventionComponentsTriggers.GOAL_SETTING.value]
-
 
 @app.task
 def intervention_component_completed(user_id: int, intervention_component_name: str):
-    phase = get_current_phase(user_id)
-    intervention_component_id = get_intervention_component_id(intervention_component_name)
-    store_intervention_component_to_db(user_id, phase.phase_id, intervention_component_id, True)
+    phase = utils.get_current_phase(user_id)
```
raar1, line 26: Can `get_current_phase()` fail (e.g. db is temporarily down or unresponsive)? If so, and this task fails, does it automatically get rescheduled (by celery?) or do we need to do this sort of error handling ourselves here?

raar1 (approved):
Good work. I just have one query about what happens when tasks fail (below)

scheduler/celery_tasks.py
wbaccinelli: Very good point. At the moment if the db is down, the task will simply fail. I will add an issue for that.

PR #54, opened 2022-07-19
Trigger reschedule future self

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/210

### DIFF
--- a/Rasa_Bot/actions/actions_rescheduling_dialog.py
+++ b/Rasa_Bot/actions/actions_rescheduling_dialog.py
@@ -131,7 +131,7 @@ class ActionRescheduleDialog(Action):
 
 def get_reschedule_options_str() -> list:
 
-    options = ["In een uur"]
+    options = ["Over 10 secondes"]
 
     current_time = datetime.datetime.now().astimezone(TIMEZONE)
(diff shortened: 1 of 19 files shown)


Rasa_Bot/actions/actions_rescheduling_dialog.py
```diff
@@ -132,4 +110,96 @@ def _is_valid_input(value):
...
+
+    # In the evening
+    elif EVENING[0] <= current_time.hour < EVENING[1]:
+        options += ["Morgenochtend, om 8:00",
+                    "Morgenmiddag, om 16:00",
+                    "Morgenavond om deze tijd"]
+
+    # In the night
+    else:
+        options += ["Vanmiddag, om 16:00",
```
bscheltinga, line 158: The night is between 0:00 and 6:00 in the morning. Then, I think the words 'vanmiddag' and 'vanavond' can be confusing and be dependent on whether the person had sleep or not.

bscheltinga (approved):
I tested it and it works! Note that you need to have the UserInterventionState populated in the db before testing.

PR #59, opened 2022-10-14
285 file sharing

wbaccinelli:
In this pull request the Rasa modifications needed to use the file sharing functionality have been implemented, together with example stories.
A detailed explanation of the workflow is reported [here](https://github.com/PerfectFit-project/virtual-coach-issues/issues/285#issuecomment-1278929847).

Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/285


Rasa_Bot/actions/actions_common.py
```diff
@@ -37,3 +37,48 @@ async def run(self, dispatcher, tracker, domain):
         logging.info("no celery error")
 
         return []
+
+
+class SendMetadata(Action):
```
raar1, line 42: Might be overkill, but a few unrelated actions could all end up together in `actions_common`, so I would add a short doctring here making clear that the `SendMetadata` relates to the `UploadFile` and `SetFilePath` actions (at least, I assume they do)

Rasa_Bot/actions/actions_common.py
```diff
@@ -37,3 +37,48 @@ async def run(self, dispatcher, tracker, domain):
...
+        return[SlotSet("uploaded_file_id", file_id)]
+
+
+class SetFilePath(Action):
+    def name(self):
+        return "action_set_file_path"
+
+    async def run(self, dispatcher, tracker, domain):
+
+        # This is hardcoded for testing. Needs to be set according to the use case
```
raar1, line 81:
```suggestion
        # TODO: This is hardcoded for testing. Needs to be set according to the use case

```

Just to make unfinished threads easier to find later ;)

Rasa_Bot/actions/actions_minimum_functional_product.py
```diff
@@ -175,3 +178,38 @@ async def run(self, dispatcher, tracker, domain):
...
+            if e['event'] == 'user':
+                user_events.append(e)
+        # we defined the metadata key name 'attachmentIds' is defined in the broker
+        received_file_ids_list = user_events[-1]['metadata']['attachmentIds']
+
+        if not self._is_valid_input(received_file_ids_list):
+            dispatcher.utter_message(response="utter_no_attachment")
+            return {"received_file_text": None}
+
+        # At this point the file ID should be saved in the DB for
```
raar1, line 206: Are we the ones sending these images, or will the user ever send images to us? Will the images ever be sensitive data that we need to be extra careful with in the db?

raar1: I guess what I'm wondering is, can anyone with the file ID fetch the file? Is it stored on Niceday servers?

Rasa_Bot/actions/actions_common.py
```diff
@@ -37,3 +37,48 @@ async def run(self, dispatcher, tracker, domain):
...
+        filepath = '/app/tst.PNG'
```
raar1, line 82: tst.PNG is a beautiful Italian Blue Italian Blue

Rasa_Bot/actions/actions_minimum_functional_product.py
```diff
@@ -175,3 +178,38 @@ async def run(self, dispatcher, tracker, domain):
...
+    def validate_received_file_text(
+            self, slot_value: Text, dispatcher: CollectingDispatcher,
+            tracker: Tracker, domain: Dict[Text, Any]) -> Dict[Text, Any]:
+        # pylint: disable=unused-argument
+        """Validate the presence of an attached file"""
+
+        # here the message metadata are retrieved
+        events = tracker.current_state()['events']
+        user_events = []
+        for e in events:
```
raar1, line 196:
```suggestion
        user_events = [e for e in events if e['event'] == 'user']
```
Would something like this be more readable? (actually maybe not...)

Rasa_Bot/actions/actions_minimum_functional_product.py
```diff
@@ -175,3 +178,38 @@ async def run(self, dispatcher, tracker, domain):
...
+            if e['event'] == 'user':
+                user_events.append(e)
+        # we defined the metadata key name 'attachmentIds' is defined in the broker
+        received_file_ids_list = user_events[-1]['metadata']['attachmentIds']
```
raar1, line 200: So the last event _always_ contains the list of received files? Is there an async problem possible where two images are sent simultaneously and the user_events list contains a mess of events for the two sends? (apologies if I've misunderstood what's happening here)

Rasa_Bot/custom_channels.py
```diff
@@ -1,33 +1,50 @@
...
 
-from rasa.core.channels.channel import InputChannel, OutputChannel, UserMessage
+from rasa.core.channels.channel import InputChannel, UserMessage, CollectingOutputChannel
 from sanic import Blueprint, response
 from sanic.request import Request
 from sanic.response import HTTPResponse
 
-from niceday_client import NicedayClient
 
 NICEDAY_API_URL = 'http://niceday_api:8080/'
```
raar1, line 11: Should this really be hard coded here? There's a `NICEDAY_API_ENDPOINT` in the `.env` configuration files that seems to be the place to set this. I'm worried this only works currently because those happen to be set the same way

Rasa_Bot/domain/domain_common.yml
```diff
@@ -1,13 +1,35 @@
...
+  - action_send_metadata
+  - action_set_file_path
 
 
 responses:
   # To refer a user to a human
   utter_refer:
-    - text: "Volgens mij heb je een probleem waar een mens je beter mee kan helpen. Ik raad je aan om contact op te nemen met iemand die je vertrouwt, zoals je huisarts."
\ No newline at end of file
+    - text: "Volgens mij heb je een probleem waar een mens je beter mee kan helpen. Ik raad je aan om contact op te nemen met iemand die je vertrouwt, zoals je huisarts."
```
raar1, line 31: Is this for when the user shares a really horrible image?

Rasa_Bot/custom_channels.py
```diff
@@ -57,18 +74,22 @@ async def health(request: Request) -> HTTPResponse:  # pylint: disable=unused-ar
         async def receive(request: Request) -> HTTPResponse:
             sender_id = request.json.get("sender")  # method to get sender_id
             text = request.json.get("message")  # method to fetch text
-
+            metadata = request.json.get("metadata")
             collector = self.get_output_channel()
             await on_new_message(
```
raar1, line 79: Does this ever timeout (or fail in some way)? Does something cause the app to retry if that happens? (this is just out of curiosity)

raar1 (approved):
Very nice work. Left some comments, some of them constructive.

Rasa_Bot/actions/actions_common.py
wbaccinelli: Definitely a good suggestion. It will help in understanding how to use such actions in the future.

Rasa_Bot/actions/actions_minimum_functional_product.py
wbaccinelli: So, the files are stored in the NiceDay server, and for accessing them the id and the access token are needed. 
The images that we will send will not contain sensitive data, and the files sent by the users will not be stored in anyway on our system, so we should be on the safe side.

Rasa_Bot/actions/actions_common.py
wbaccinelli: not a random choice XD

Rasa_Bot/actions/actions_minimum_functional_product.py
wbaccinelli: This is a validation action that runs on user text input and the new message structure created in the custom channels always contains the metadata (even if the files list is empty). So, in theory we always get this. But theory and reality do not always match. It is a good idea to surround this with some verification.

This action runs on a single message, so the is no possibility to mess with this action.

Rasa_Bot/custom_channels.py
wbaccinelli: good point. I will change this to use the .env file as the rest of the components.

Rasa_Bot/custom_channels.py
wbaccinelli: This contacts the Rasa server, that has a timeout. But if it fails for whatever reason, at the moment we do nothing, you are right

wbaccinelli: This contacts the Rasa server, that has a timeout. But if it fails for whatever reason, at the moment we do nothing, you are right

PR #60, opened 2022-11-09
get five highest rated activities

wbaccinelli:
Changed the rasa_action to retrieve the first aid kit, in order to get only the 5 activities with the highest rating


raar1 (approved):
Looks good

PR #61, opened 2022-11-16
290 general activity

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/290

The general activity dialog has been implemented in this PR.


Rasa_Bot/data/stories_general_activity.yml
```diff
@@ -0,0 +1,163 @@
...
+  steps:
+  - checkpoint: choose_activity_checkpoint
+  - action: check_who_decides
+  - slot_was_set:
+    - who_decides_slot: user
+  - action: get_three_random_activities
+  - action: utter_general_activity_choose_next_activity
+  - action: general_activity_next_activity_form
+  - active_loop: general_activity_next_activity_form
+  - active_loop: null
```
raar1, line 147: Sorry for my own ignorance but what is the purpose of the `active_loop: null` ?

raar1 (approved):
OK this looks good to me, although I'm not so familiar with the Rasa dialogs so I can't review in depth. It looks like a huge amount has been added here. Some of the yaml files are getting quite large so I suppose it is tricky to keep track when adding more content? (I haven't tried this myself, so maybe it's easy enough).

I was intrigued by the mysterious "ordered-chateau" file

Rasa_Bot/data/stories_general_activity.yml
wbaccinelli: The `active_loop: null` is required by the syntax in Rasa in order to close the forms. Rasa...

wbaccinelli:
Yes, the content is becoming really huge. We are using different files for each conversation, to contain the complexity, but it will be tricky still. 

The "ordered-chateau" is the rasa trained mode. The names are randomly assigned and are very funny.

PR #62, opened 2022-12-07
External Triggering

BendikC:
Fixed the external triggering by including two input channels, one for the triggering and one for the regular conversation flow. Each input channel has a corresponding output channel and they function as intended.


wbaccinelli (approved):
Good job!

PR #64, opened 2022-12-14
Relapse dialog

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/291

The Relapse dialog has been completely implemented. The graph sharing functionality will be integrated in [this issue](https://github.com/PerfectFit-project/virtual-coach-issues/issues/288)

The interconnection with the goal-setting and medication dialogs will be implemented when those dialogs will be ready.


Rasa_Bot/actions/actions_execution_dialogs.py
```diff
@@ -42,3 +42,13 @@ def name(self):
...
         return [SlotSet("current_intervention_component",
                         ExecutionInterventionComponents.DAILY_REFLECTION)]
+
+
+class SetSlotRelapseDialog(Action):
+    def name(self):
+        return "action_set_slot_relapse_dialog"
+
+    async def run(self, dispatcher, tracker, domain):
+        print(ExecutionInterventionComponents.RELAPSE_DIALOG)
```
raar1, line 52: Do we have a proper logging system yet, or do we still just rely on print statements?

Rasa_Bot/actions/actions_first_aid_kit.py
```diff
@@ -1,10 +1,55 @@
...
+
+class ActionStartFak(Action):
+    """Action to run the First Aid Kit from external trigger"""
+
+    def name(self):
+        return "action_start_fak"
+
+    async def run(self, dispatcher, tracker, domain):
+
+        user_id = int(tracker.current_state()['sender_id'])  # retrieve userID
```
raar1, line 23: Is sender_id a string? Is it possible this conversion to int ever fails?

Rasa_Bot/actions/actions_execution_dialogs.py
raar1: Maybe use logging.info() ?

Rasa_Bot/actions/actions_first_aid_kit.py
```diff
@@ -1,10 +1,55 @@
...
+        print("CELERY INTENT")
```
raar1, line 24: logging.info()?

Rasa_Bot/actions/actions_future_self_dialog.py
```diff
@@ -65,7 +71,10 @@ def name(self):
     async def run(self, dispatcher, tracker, domain):
         answer = tracker.get_slot("why_picked_words")
         user_id = tracker.current_state()['sender_id']
```
raar1, line 73: So here user_id does not need to have been converted to int

Rasa_Bot/actions/helper.py
```diff
@@ -2,42 +2,181 @@
...
+            answer_value: the value chosen by the user
+
+        Returns:
+                nothing
+
+    """
+    session = get_db_session(db_url=DATABASE_URL)  # Create session object to connect db
+    selected = session.query(Users).filter_by(nicedayuid=user_id).one()
+
+    answer_id = answer_value + question_id * 100
```
raar1, line 33: What is happening here? Maybe a short comment would help

Rasa_Bot/actions/helper.py
```diff
@@ -2,42 +2,181 @@
...
+        )
+        .filter(
+            InterventionComponents.intervention_component_name == intervention_component
+        )
+        .all()
+    )
+
+    session.add(UserInterventionState(
+        users_nicedayuid=user_id,
+        intervention_phase_id=phases[0].phase_id,
```
raar1, line 158: Is there a risk here that phases and components are empty? (i.e. [0] indexing will fail)
Maybe check for this earlier to fail more gracefully?

Rasa_Bot/actions/helper.py
```diff
@@ -55,21 +194,66 @@ def get_intervention_component_id(intervention_component_name: str) -> int:
...
+
+    """
     events_bot = []
 
     for event in events:
         if event['event'] == 'bot':
             events_bot.append(event)
 
-    if len(events_bot) != 0:
+    if len(events_bot) != 0 and 'utter_action' in events_bot[-1]['metadata']:
```
raar1, line 213: If `events_bot` is not empty is it guaranteed to have a 'metadata' key in the last element?

Rasa_Bot/data/rules/rules.yml
```diff
@@ -0,0 +1,64 @@
...
+  - action: utter_refer
+  - action: utter_central_mode_options
+  - action: action_end_dialog
+
+- rule: Ask for foreseen smoking HRS whenever this is externally triggered (by scheduler)
+  steps:
+  - intent: EXTERNAL_trigger_ask_foreseen_hrs
+  - action: utter_ask_foreseen_hrs
+
+#- rule:  Return the number of smoked cigarettes when the user asks for it
```
raar1, line 25: Are these just temporarily commented out?

Rasa_Bot/domain/domain_generic.yml
```diff
@@ -1,5 +1,5 @@
 session_config:
-  session_expiration_time: 5 # these are minutes
+  session_expiration_time: 60 # these are minutes
```
raar1, line 2: What happens when the session expires? Is this something we should be careful about (i.e. test) regarding the rest of the code to make it robust against this eventuality? Also, are there several sessions (per user?) or only one?

raar1 (approved):
Good job man, this is a lot of work. This is getting to be quite complex so we'll need to do a lot of quality assurance/testing once all these dialogs are implemented... :)

Rasa_Bot/actions/actions_execution_dialogs.py
wbaccinelli: This is leftover from debugging, I'm deleting that. We definitely need to implement a proper logging system

Rasa_Bot/actions/actions_first_aid_kit.py
wbaccinelli: sender_id is a string, because Rasa allows all sorts of IDs. In our case this is the ID we get from NiceDay, which is a progressive number (so far). We need the conversion to store it in the DB, where it is defined as an int. Would it be safer to have everything as a string?

Rasa_Bot/actions/helper.py
wbaccinelli: The answers to the closed questions are pre-defined and initialized in the DB. To have a unique known ID for the answers that we can use to store the user's response, it is assigned by combining the question id and the value of the answer (it is always a number). A comment is really needed, I am adding it.

Rasa_Bot/actions/helper.py
wbaccinelli: Phases and components are initialized in the DB, but if a non existing phase or component is provided as argument, this will fail. I am adding a control

Rasa_Bot/actions/helper.py
wbaccinelli: right, it is good to add the check

Rasa_Bot/data/rules/rules.yml
wbaccinelli: These are test rules. I should leave them uncommented unless we get rid of them

Rasa_Bot/domain/domain_generic.yml
wbaccinelli: There is one session per user, and when it expires the conversation is reset. This means that the user has 60 minutes to answer the last question of the VC, then the dialog will be over.

PR #67, opened 2023-01-13
310 relapse goal setting medication

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/310


bscheltinga (approved):
Hi Walter, nice work! It looks good, I have a few suggestions and think caught one bug (but not from this PA).

bscheltinga:
I tested it and it works!

PR #68, opened 2023-01-20
288 share and create graphs

BendikC:
Completed the issue, created the graphs, fixed the problem with the wrong retrieval from the database, so that part is fine now. 

Only thing to note is that I delete the file again in the action "send_metadata" as that is the last place it is used, because I think the only time we will used send_metadata, is when we upload something ourselves and send it. If this is not the case, then the deletion needs to be in a separate action.


Rasa_Bot/actions/helper.py
```diff
@@ -264,6 +263,109 @@ def get_random_activities(avoid_activity_id: int, number_of_activities: int
...
+    result = []
+    for closed_answer_option in closed_answer_options:
+        amount_of_answers = 0
+        closed_answer = closed_answer_option
+
+        for answer in answers:
+            if answer.closed_answers_id == closed_answer.closed_answers_id:
+                amount_of_answers += 1
+
+        result.append(amount_of_answers)
```
wbaccinelli, line 365: You can use nested comprehension lists to do this in a more performant way.

```suggestion
       result = [
        len(
            [answer for answer in answers if answer.closed_answers_id == closed_answer.closed_answers_id]
        ) for
              closed_answer in closed_answer_options
    ]
```

Rasa_Bot/actions/helper.py
```diff
@@ -280,4 +382,59 @@ def week_day_to_numerical_form(week_day):
...
+        closed_answer_options = get_all_closed_answers(question_ids_subset[0])
+
+        for question_id in question_ids_subset:
+            closed_answers = get_all_closed_answers(question_id)
+            answers = get_closed_answers(user_id, question_id)
+            data.append(count_answers(answers, closed_answers))
+
+        answer_descriptions = []
+        for answer in closed_answer_options:
+            answer_descriptions.append(answer.answer_description)
```
wbaccinelli, line 433:
```suggestion
        closed_answer_options = get_all_closed_answers(question_ids_subset[0])

        data = [count_answers(get_closed_answers(user_id, question_id), get_all_closed_answers(question_id)) for
                question_id in question_ids_subset]

        answer_descriptions = [answer.answer_description for answer in closed_answer_options]
```

wbaccinelli (approved):
Good job! I left some suggestions.
As a general comment, it would be good to use type annotation when defining new functions in the helper. It helps using them correctly.

Rasa_Bot/actions/actions_common.py
```diff
@@ -60,6 +61,8 @@ async def run(self, dispatcher, tracker, domain):
             json_message={"text": "image",
```
wbaccinelli, line 61: We should use a different text. Either an empty one or something like "Click on the image to see the full picture", because the preview is small.

PR #69, opened 2023-01-23
Testim

nelealbers:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/315.

### DIFF
--- a/Rasa_Bot/actions/actions_goal_setting_dialog.py
+++ b/Rasa_Bot/actions/actions_goal_setting_dialog.py
@@ -1,10 +1,12 @@
 """
 Contains custom actions related to the relapse dialogs
 """
+from virtual_coach_db.dbschema.models import (Testimonials, Users)
 from virtual_coach_db.helper import ExecutionInterventionComponents, PreparationInterventionComponents
+from virtual_coach_db.helper.helper_functions import get_db_session
 
 from . import validator
-from .definitions import TIMEZONE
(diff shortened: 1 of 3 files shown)


nelealbers:
@bscheltinga The prospector complains about something in your code, but I am not sure what the logic is supposed to be so I can't fix that.

Rasa_Bot/actions/actions_goal_setting_dialog.py
```diff
@@ -26,6 +29,78 @@ async def run(self, dispatcher, tracker, domain):
 
         return [SlotSet('first_possible_quit_date', first_date),
                 SlotSet('last_possible_quit_date', last_date)]
+    
+    
+def goal_setting_testimonial_model_output(t, user_se, user_godin, user_c1,
+                                          user_c3):
```
wbaccinelli, line 35: It would be better to add type annotation to this function, and to add parameters and return values explanation in the docstring function explanation. These small improvements help the understanding of the function.

wbaccinelli (approved):
Nice job!

Rasa_Bot/actions/actions_goal_setting_dialog.py
nelealbers: Good point!

bscheltinga (approved):
Looks good! I also tested it and it works on my machine!

PR #70, opened 2023-01-25
313 preparation phase dialog

7lopol7:
Fix https://github.com/PerfectFit-project/virtual-coach-issues/issues/313

Added the preparation phase dialog

### DIFF
--- a/Rasa_Bot/actions/actions_preparation_phase.py
+++ b/Rasa_Bot/actions/actions_preparation_phase.py
@@ -1,12 +1,12 @@
 from rasa_sdk import Action, Tracker
 from rasa_sdk.events import SlotSet
-from virtual_coach_db.helper.definitions import PreviousDialog
+from virtual_coach_db.helper.definitions import DialogExpectedDuration
 from rasa_sdk.executor import CollectingDispatcher
 from rasa_sdk.forms import FormValidationAction
 from typing import Text, Dict, Any
 from celery import Celery
 from . import validator
-from .definitions import REDIS_URL, MORNING, AFTERNOON, EVENING, TIMEZONE
+from .definitions import REDIS_URL, MORNING, AFTERNOON, TIMEZONE
 from .helper import get_latest_bot_utterance
 from .actions_rescheduling_dialog import get_reschedule_date
 import datetime
@@ -23,7 +23,7 @@ class SetSlotPreviousDialog(Action):
 
     async def run(self, dispatcher, tracker, domain):
         return [SlotSet("expected_next_time_interval",
-                        PreviousDialog.DIALOG1)]
+                        DialogExpectedDuration.DIALOG1)]
 
 
 class ExpectedTimeNextPart(Action):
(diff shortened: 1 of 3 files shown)


Rasa_Bot/actions/actions_preparation_phase.py
```diff
@@ -0,0 +1,137 @@
...
+class SetSlotPreviousDialog(Action):
+    """ this is an example for setting the time interval of the previous dialog,
+     it should be incorporated in the regular method of setting the dialog slot"""
+
+    def name(self):
+        return "action_set_slot_previous_dialog"
+
+    async def run(self, dispatcher, tracker, domain):
+        return [SlotSet("expected_next_time_interval",
+                        PreviousDialog.DIALOG1)]
```
wbaccinelli, line 26: Is a separate action with a new slot needed for this? Can read the time span value directly in `action_expected_time_next_part`?

Rasa_Bot/actions/actions_preparation_phase.py
```diff
@@ -0,0 +1,137 @@
+from rasa_sdk import Action, Tracker
+from rasa_sdk.events import SlotSet
+from virtual_coach_db.helper.definitions import PreviousDialog
```
wbaccinelli, line 3: `PreviousDialog` is not in the main branch of `virtual_coach_db`. Be sure to open a PR for this. Also, maybe a naming like `DialogExpectedDuration` might be more explicative?

Rasa_Bot/data/rules_preparation_phase.yml
```diff
@@ -0,0 +1,50 @@
...
+
+  condition:
+  - active_loop: next_part_now_or_later_form
+  steps:
+    - action: next_part_now_or_later_form
+    - active_loop: null
+    - slot_was_set:
+        - now_or_later: 1
+    - action: utter_lets_start
+    - action: action_end_dialog
```
wbaccinelli, line 25: Instead of ending the dialog here we should trigger the execution of the new dialog. An action can be implemented that gets the `current_intervention_component` slot, which is set at when the dialog is proposed, and send the external trigger consequently.

wbaccinelli (approved):
Nice job! I left some comments for making the flow a bit more consistent.
Also, fix the issues suggested by the Python CI before merging

Rasa_Bot/actions/actions_preparation_phase.py
7lopol7: Yes can I then use tracker.get_slot('current_intervention_component') to get the dialog before the current one? And let the db helper method return a timespan based on this slot value?

Rasa_Bot/actions/actions_preparation_phase.py
7lopol7: Ah yes will do!

Rasa_Bot/actions/actions_preparation_phase.py
wbaccinelli: Yes, exactly

Rasa_Bot/data/rules_preparation_phase.yml
7lopol7: should this trigger be sent via Celery or can it done directly in this action?

wbaccinelli: This can be done in both ways. I would implement it with Celery now, for clarity.

7lopol7:
@wbaccinelli I implemented the changes you suggested, could you take a look at them?

Rasa_Bot/actions/actions_preparation_phase.py
```diff
@@ -80,6 +69,16 @@ def validate_chosen_daypart(
...
 
+class StartNextDialog(Action):
+    """set trigger for next dialog"""
+    def name(self) -> Text:
+        return "action_start_next_dialog"
+
+    async def run(self, dispatcher, tracker, domain):
+        user_id = tracker.current_state()['sender_id']
+        nextDialog = tracker.get_slot('current_intervention_component')
+        celery.send_task('celery_tasks.trigger_intervention_component', (user_id, nextDialog))
```
wbaccinelli, line 80: `celery_tasks.trigger_intervention_component` requires the trigger rather than the name of the component to trigger. Probably it would be better to make the scheduler use the name, but at the moment we have to use the trigger. Can you implement here the logic to get the trigger from the `current_intervention_component`? We have the definitions for the triggers and the components in the DB

wbaccinelli (approved):

Rasa_Bot/actions/actions_preparation_phase.py
7lopol7: Yess will do!

7lopol7: Yes will do!

7lopol7: Can these components be both preparation and execution intervention components? Or are they always preparation intervention components?

wbaccinelli: They will be preparation

PR #71, opened 2023-02-07
Add a sensor api to update the step count

kkkksu:
example post
```
curl -X 'POST' \
  'http://172.23.97.218:8080/sensor/step' \
  -H 'accept: */*' \
  -H 'Content-Type: application/json' \
  -d '{
  "localTime": "2023-01-19T09:00:00.000",
  "timezone": "Europe/Amsterdam",
  "token": "345ddcad9741fd8a8aff4161fd156fb2307d00210c9e0b77963a25e09e488318",
  "user": "9o0xZ0JGRauKnlY31dYJjA",
  "utcTime": 1674115200000,
  "value": 14
}'
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,6 +85,11 @@ ipython_config.py
 # pyenv
 .python-version
 
+sensor_api/bin
+sensor_api/lib
+sensor_api/share
+sensor_api/lib64
+sensor_api/pyvenv.cfg
 # pipenv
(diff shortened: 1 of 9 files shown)


wbaccinelli (approved):
Nice job! I wonder whether it may be better to have some tasks queuing system, in order to be sure that we can manage the uploads without losing the data.

wbaccinelli:
Outdated

PR #74, opened 2023-03-01
311 time structure

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/311

In this PR the state machine for the administration of the intervention has been implemented. Three main components are used:

### StateMachine
The `StateMachine` class is defined in the 'state_machine.py' script. It is in charge of keeping track of the current intervention state. It also receive different types of events as defined in the 'Event' class, and decides which function of the state to activate depending on the event type. 
6 events have been defined:
- DIALOG_COMPLETED: a dialog has been completed in Rasa by the user. The completion is tracked and the logic to manage it is activated in the specific state
- DIALOG_RESCHEDULED: a dialog has been rescheduled. The specific state manages the actions to take
- DIALOG_STARTED: a dialog has started. The starting time and the event are tracked
- NEW_DAY: a new day has started. Some of the intervention states take different actions depending on the time dimension.
- USER_TRIGGER = a dialog has been triggered by the user in Rasa. The specific state manages the actions to be taken.

Each user is associated with the instance of a StateMachine.
The state machine is coupled with the `user_state_machine` table on the DB. Every time an event occurs, the state machine is populated with the data stored in the DB, the event is consumed, and the new parameters of the state machine are updated on the DB.

### Controller
In the `controller.py` each possible state is fully defined. Each state is an implementation of the abstract `State` class defined in `state.py`. Each state implements a different logic for managing the events sent by the StateMachine, and defines the rules for changing state.

### celery_tasks
In `celery_tasks.py` the possible celery tasks are defined and implemented. These tasks can be triggered both from Rasa and the states of the state machine.


Rasa_Bot/actions/requirements-actions.txt
```diff
@@ -6,6 +6,6 @@ python-dateutil==2.8.1
 python-dotenv
 redis
 sqlalchemy
-git+https://github.com/PerfectFit-project/virtual-coach-db@v0.1.10
+git+https://github.com/PerfectFit-project/virtual-coach-db@311-time-structure
```
raar1, line 9: I suppose this will need to be updated once the PR is merged

scheduler/state_machine/const.py
```diff
@@ -0,0 +1,25 @@
...
+EXECUTION_DURATION = 12 * 7  # 12 weeks
+#
+
+REDIS_URL = os.getenv('REDIS_URL')
+DATABASE_URL = os.getenv('DATABASE_URL')
+TRIGGER_COMPONENT = 'celery_tasks.trigger_intervention_component'
+SCHEDULE_TRIGGER_COMPONENT = 'celery_tasks.trigger_scheduled_intervention_component'
+TIMEZONE = tz.gettz("Europe/Amsterdam")
+
+# dialog states definitions
```
raar1, line 22: I suppose these could be enums too?

scheduler/state_machine/state_machine.py
```diff
@@ -0,0 +1,148 @@
...
+            # in this case the descriptor is a tuple, where 0 is
+            # the dialog and 1 the new date
+            self.state.on_dialog_rescheduled(event.Descriptor[0], event.Descriptor[1])
+
+        if event.EventType == EventEnum.DIALOG_STARTED:
+            logging.info('Dialog started event received %s ', event.Descriptor)
+            # in this case we track that a dialog is running
+            self.dialog_state.set_to_running(event.Descriptor)
+
+        elif event.EventType == EventEnum.NEW_DAY:
```
raar1, line 104: Only this if is an "else if" - should all the others not use `elif` too? It seems that there can only be a unique value of `EventType` (and changes in dialog_state don't modify the event, I assume)

scheduler/celery_tasks.py
```diff
@@ -1,195 +1,201 @@
...
+    state_machines = get_all_fsm()
+
+    for fsm in state_machines:
+        dialog_state = get_dialog_state(fsm)
+
+        if dialog_state == EXPIRED:
+            dialog = fsm.dialog_state.get_current_dialog()
+            
+            # the dialog is idle now
+            fms.dialog_state.set_to_idle()
```
raar1, line 60: This should be `fsm` I think?

scheduler/state_machine/controller.py
```diff
@@ -0,0 +1,573 @@
...
+
+        elif dialog == Components.WEEKLY_REFLECTION:
+            logging.info('Weekly reflection completed')
+
+            week = get_execution_week(user_id=self.user_id)
+            
+            # on the first week, the execution of the general activity dialog 
+            # has to be planned for week 2
+            if week == 1:
+              schedule_next_execution(user_id=self.user_id,
```
raar1, line 345:
```suggestion
                schedule_next_execution(user_id=self.user_id,

```
Fix indentation

raar1 (approved):

scheduler/celery_tasks.py
```diff
@@ -1,195 +1,197 @@
...
-        completed=False,
-        last_time=datetime.now().astimezone(TIMEZONE),
-        last_part=0,
-        next_planned_date=None,
-        task_uuid=None
-    )
 
-    utils.store_intervention_component_to_db(state)
+@app.task(bind=True)
+def notify_new_day(self, current_date: date):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
wbaccinelli: Upon the service restart, the scheduled tasks are automatically executed by celery. This is not going to happen in case we it crashed so badly that we have to rebuild it. Do you think we should manage also this?

scheduler/celery_tasks.py
```diff
@@ -1,195 +1,197 @@
...
+        trigger: the intent to be sent
+    """
+    name = get_component_name(trigger)
+    send_fsm_event(user_id, Event(EventEnum.DIALOG_STARTED, name))
+
     endpoint = f'http://rasa_server:5005/conversations/{user_id}/trigger_intent'
     headers = {'Content-Type': 'application/json'}
     params = {'output_channel': 'niceday_trigger_input_channel'}
     data = '{"name": "' + trigger + '" }'
     requests.post(endpoint, headers=headers, params=params, data=data, timeout=60)
```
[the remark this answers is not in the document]
wbaccinelli: You're right, I fixed it

scheduler/state_machine/state_machine.py
wbaccinelli: Yes, probably I copy/pasted the wrong line over and over 😅

scheduler/celery_tasks.py
wbaccinelli: yes, stupid typos

PR #83, opened 2023-04-19
fixed some flows

wbaccinelli:


Rasa_Bot/data/nlu.yml
```diff
@@ -167,38 +107,10 @@ nlu:
   examples: |
     - "EXTERNAL_rescheduling_preparation_phase"
 
-- intent: relapse_dialog
```
bscheltinga, line 170: Don't we need an intent for this dialog as it is user triggered as well?

bscheltinga: ahaha I just figured out and wanted to say that :)

bscheltinga (approved):

PR #84, opened 2023-04-24
338 review general activity

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/338

It works in combination with the changes implemented in [this PR](https://github.com/PerfectFit-project/virtual-coach-db/pull/58)


Rasa_Bot/actions/helper.py
```diff
@@ -340,6 +341,117 @@ def store_user_intervention_state(user_id: int, intervention_component: str, pha
...
+        session.query(
+            UserStateMachine
+        )
+        .filter(
+            UserStateMachine.users_nicedayuid == user_id
+        )
+        .all()
+    )
+
+    return user_fsm[0].state
```
raar1, line 391: This seems to be a common pattern in the code here, but instead of using `.all()` followed by [0] indexing, you could perhaps use the `.first()` sqlalchemy function instead. (see <https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.first>)

I'm not sure how many results you expect of each of these queries (perhaps only one anyway?) but if there can be lots then maybe using `.first()` can improve performance

Rasa_Bot/actions/helper.py
```diff
@@ -429,6 +541,90 @@ def get_random_activities(avoid_activity_id: int, number_of_activities: int
...
+            # have rescheduled the dialog, so we need to take that into account. In case
+            # the activity has been already performed, it is removed form the mandatory list
+            if "mandatory" in phase:
+                for mandatory_time in phase["mandatory"]:
+                    if curr_time >= mandatory_time:
+                        mandatory.append(resource["resource_id"])
+
+    mandatory_ids = []
+    # check if the mandatory activities have been already performed
+    if len(mandatory) > 0:
```
raar1, line 596: I don't think you need this `if` statement. If there are no elements in mandatory then the for loop will not execute anyway

scheduler/state_machine/controller.py
```diff
@@ -149,6 +149,14 @@ def run(self):
 
     def on_new_day(self, current_date: date):
         logging.info('current date: %s', current_date)
+        # at day 7 activity C2.9 has to be proposed
+
+        start_date = get_start_date(self.user_id)
+
+        if(current_date - start_date).days >= 7:
```
raar1, line 156: Perhaps this `7` should be represented by a named constant (e.g. `ACTIVITY_C2_9_DAY_TRIGGER` or something)?

raar1 (approved):
Looks good to me. As before, I am not very useful for reviewing the text in the yml files, but I have left a couple of suggestions for improvements :)

PR #85, opened 2023-05-02
Restart dialogs

nelealbers:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/296 for the goal-setting dialog.


nelealbers:
@wbaccinelli , could you please take a look what I am missing in actions_idle_commands.py to let the state machine figure out whether there is an uncompleted dialog after the user types "verder"?

Rasa_Bot/actions/helper.py
```diff
@@ -95,8 +100,102 @@ def compute_preferred_time(tracker):
...
+    session = get_db_session(db_url=DATABASE_URL)
+
+    selected = (
+        session.query(
+            UserInterventionState
+        ).order_by(UserInterventionState.last_time.desc())
+        .filter(
+            UserInterventionState.users_nicedayuid == user_id,
+            UserInterventionState.intervention_component_id == intervention_component_id
+        )
```
wbaccinelli, line 116: in the table there is also the `completed` field that we can check to filter out the results.

```suggestion
            UserInterventionState.users_nicedayuid == user_id,
            UserInterventionState.intervention_component_id == intervention_component_id
            UserInterventionState.completed == False
```

wbaccinelli:
Moving everythting to the FSM was too complicated, so the actions will check the presence of uncompleted dialog, and the FSM will trigger the right one. In the FMS there is also an additional layer of checking, in case the user types 'verder' (which triggers the rule for continuing the dialog in any case) when this is not allowed.

wbaccinelli (approved):
It works perfectly! I added the mechanism for the menu selection and for starting the dialog to be continued. Let me know if it looks good to you

PR #87, opened 2023-05-17
341 goal imp

nelealbers:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/341 using conditional response variations.

### DIFF
--- a/Rasa_Bot/actions/actions_goal_setting_dialog.py
+++ b/Rasa_Bot/actions/actions_goal_setting_dialog.py
@@ -496,7 +496,7 @@ class ValidateWhyPaGoalImportantForm(FormValidationAction):
         if last_utterance != 'utter_ask_why_pa_goal_important_values':
             return {"why_pa_goal_important_values": None}
 
-        if not validator.validate_list(value, 1, 6):
+        if not validator.validate_number_in_range_response(1, 6, value):
             dispatcher.utter_message(response="utter_did_not_understand")
             dispatcher.utter_message(response="utter_please_answer_numbers")
             return {"why_pa_goal_important_values": None}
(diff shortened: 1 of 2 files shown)


nelealbers:
@bscheltinga Ah I see, that is not what we had in Beyza's dialog that was the basis of the dialog, so I did not even think that it would be an option. In that case conditional response variations won't work.

bscheltinga:
Yes, I figured out that a conditional response would not work. I think we have two options now:
1. Keep it as it is
2. Make the response generic

With generic response I mean the following, currently we have:
>   utter_ask_why_pa_goal_important_nuance:
>     - text: "Ik kan me voorstellen dat jouw {why_pa_goal_important_values} belangrijk voor je is/zijn. Kun je me iets
>             meer vertellen over waarom dit doel belangrijk is voor jouw {why_pa_goal_important_values}?"

We can change that into:

>   utter_ask_why_pa_goal_important_nuance:
>     - text: "Ik kan me voorstellen dat dit belangrijk voor je is. Kun je me vertellen waarom je dit hebt gekozen?"

nelealbers:
@bscheltinga I was at a conference so just looked at it now. I think the actual problem is that the implementation assumes that people can type multiple numbers for their goal importance. But this is not the case - the user should just type one. The goal importance comes from Beyza's dialog, where people can only choose one. And the excel-file for the goal-setting dialog also does not say anything about being able to choose multiple answers.

bscheltinga:
@nelealbers I fixed the validation and added the sentence that they need to choose the option which is most important for them: "Kies de optie die voor jouw het belangrijkste is."

bscheltinga:
@nelealbers you're right. I retrained the model.

nelealbers:
@bscheltinga Nice, I merged it!

PR #91, opened 2023-06-02
changed dialog checking and new day to periodic tasks

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/344

### DIFF
--- a/scheduler/celery_tasks.py
+++ b/scheduler/celery_tasks.py
@@ -22,7 +22,7 @@ def setup_periodic_tasks(sender, **kwargs):  # pylint: disable=unused-argument
     When celery is ready, the watchdogs for the new day notification
     and for the dialogs status check are started
     """
-    sender.add_periodic_task(crontab(hour=13, minute=37), notify_new_day.s(datetime.today()))
+    sender.add_periodic_task(crontab(hour=00, minute=00), notify_new_day.s(datetime.today()))
     sender.add_periodic_task(MAXIMUM_DIALOG_DURATION, check_dialogs_status.s())
 
 
--- a/scheduler/celery_tasks.py
+++ b/scheduler/celery_tasks.py
@@ -1,8 +1,7 @@
 import logging
-import os
 import requests
 from celery import Celery
-from celery.signals import worker_ready
+from celery.schedules import crontab
 from datetime import date, datetime, timedelta
 from state_machine.state_machine import EventEnum, Event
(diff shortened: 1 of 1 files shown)


scheduler/celery_tasks.py
```diff
@@ -68,16 +64,12 @@ def check_dialogs_status(self):  # pylint: disable=unused-argument
             fsm.dialog_state.set_to_idle()
             save_state_machine_to_db(fsm)
 
-            next_day = datetime.now().replace(hour=10, minute=00) + timedelta(days=1)
+            next_day = datetime.now().replace(hour=00, minute=00) + timedelta(days=1)
```
raar1, line 67: The new day used to start at 10am?

scheduler/celery_tasks.py
```diff
@@ -16,18 +15,15 @@
...
+def setup_periodic_tasks(sender, **kwargs):  # pylint: disable=unused-argument
     """
     When celery is ready, the watchdogs for the new day notification
     and for the dialogs status check are started
     """
-    notify_new_day.apply_async(args=[datetime.today()])
-    ## TODO: uncomment and solve issue
-    # check_dialogs_status.apply_async()
+    sender.add_periodic_task(crontab(hour=00, minute=00), notify_new_day.s(datetime.today()))
+    sender.add_periodic_task(MAXIMUM_DIALOG_DURATION, check_dialogs_status.s())
```
raar1, line 26: Something I'm wondering about - I think we configured Celery to use the redis key-value store for storing its tasks. If the current docker-compose setup goes down, the contents of that store are lost, so on restarting the app it should be completely clear (no celery tasks set at all). So the app initially sets up every celery task it needs just from reading the db. Is this the correct way to understand it?

raar1 (approved):
OK this looks cleaner than before :) I assume using celery's periodic tasks is more robust than the async stuff from before?

scheduler/celery_tasks.py
wbaccinelli: This is a very good point, and it might be a problem if we want, for example, to update the system. At the moment the app sets up just recurrent tasks for checks that need to be done periodically, but it doesn't restore the scheduled tasks that have not been executed. This information is in the DB, so  we "just" need to read the info from there. I opening an issue for that.

scheduler/celery_tasks.py
wbaccinelli: Yes, but it didn't make a lot of sense. I mean, I am not yet so lucky to be in the position of denying the existence of the early morning hours XD

PR #92, opened 2023-06-05
248 oboarding

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/91
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/248

Celery checks programmatically if a new connection request has been sent to the therapist. In case a new request is found and the user is not yet registered in the DB, the request is accepted. Upon the request acceptance, the needed information are stored in the DB and the intervention is started.

### DIFF
--- a/scheduler/celery_utils.py
+++ b/scheduler/celery_utils.py
@@ -37,16 +37,18 @@ def create_new_user(user_id: int):
         user_id: the ID of the user
     """
 
-    user_exists = check_if_user_exists(user_id)
+    # if a user exists already, print an error message
+    if check_if_user_exists(user_id):
+        logging.warning('User profile already in the DB')
+        return
 
-    if not user_exists:
-        new_user_profile = create_new_user_profile(user_id)
-        save_user_to_db(new_user_profile)
-        new_fsm = create_new_user_fsm(user_id)
-        save_state_machine_to_db(new_fsm)
+    new_user_profile = create_new_user_profile(user_id)
+    new_fsm = create_new_user_fsm(user_id)
 
-    else:
-        logging.warning('The user already exists in the database')
+    session = get_db_session(DATABASE_URL)
+    session.merge(new_user_profile)
+    session.merge(new_fsm)
+    session.commit()
 
 
 def create_new_user_profile(user_id: int) -> Users:
@@ -63,7 +65,7 @@ def create_new_user_profile(user_id: int) -> Users:
     return user
 
 
-def create_new_user_fsm(user_id: int) -> StateMachine:
+def create_new_user_fsm(user_id: int) -> UserStateMachine:
     """
     Creates a new StateMachine for the user specified.
     Args:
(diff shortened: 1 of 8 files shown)


scheduler/celery_utils.py
```diff
@@ -30,9 +30,42 @@ def check_if_user_exists(user_id: int) -> bool:
...
+def create_new_user(user_id: int):
+    """
+    Initialize the DB for a new user.
+    Args:
+        user_id: the ID of the user
+    """
+
+    user_exists = check_if_user_exists(user_id)
+
+    if not user_exists:
```
raar1, line 42: Can there ever be a race condition with this? How many threads does celery use by default (>1?) for its workers? I'm wondering if, for some reason, the app was spammed with loads of connection requests for the same userid, would multiple celery threads be working on adding the same user simultaneously? In such a case the user existence check would be subject to a race condition and everything in the `if not user_exists:` block could end up being executed even if the user exists. (Also, would this cause big problems if it happened?)

scheduler/celery_utils.py
```diff
@@ -30,9 +30,42 @@ def check_if_user_exists(user_id: int) -> bool:
...
+        new_user_profile = create_new_user_profile(user_id)
+        save_user_to_db(new_user_profile)
+        new_fsm = create_new_user_fsm(user_id)
+        save_state_machine_to_db(new_fsm)
```
raar1, line 46: If something kills the celery worker half way through executing this function, so that `save_user_to_db` has run, but not yet `save_state_machine_to_db`, would that be a serious problem? Would the app be able to recover from it, or reinitiate the onboarding?

If not, then one solution could be adding an `onboarding_complete` flag to the user entry in the db, that is only set to true once all onboarding info has been created and written to the db. Then that status could be checked before certain operations to recover from partial onboarding. Perhaps you already have something that effectively does this?

raar1: Actually, my preference would be to write both the new user and the new state_machine to the db simultaneously (i.e. in the same `session.commit()`, if that's possible. That way it is guaranteed that the db will only have either a completely onboarded user, or no user.

scheduler/state_machine/const.py
```diff
@@ -4,6 +4,9 @@
 # maximum duration of the dialog in seconds
 MAXIMUM_DIALOG_DURATION = 90*60
 
+# time interval in seconds for checking for new connection requests
+INVITES_CHECK_INTERVAL = 30*60
```
raar1, line 8: I suppose users somehow receive an email or other notification once they have been onboarded? Or does the chat just stay unresponsive for 30 mins?

raar1 (approved):
This looks like it should work, but I have left a few comments that I think you should address if they're not already. Apologies in advance for my paranoia about race conditions... :neutral_face:

scheduler/celery_utils.py
wbaccinelli: Good point, I will put all together

scheduler/state_machine/const.py
wbaccinelli: So, they are not able to chat with the coach if the onboarding has not been completed. If the enter the app they will see "Connection pending". This interval can also be reduced, but I didn't want to run a task every few seconds.

scheduler/celery_utils.py
raar1: No I think it's good to run as many tasks in parallel as possible. Also yes, the loop will indeed run only in one task, but if there were two identical tasks both executing this then they will each be executing their own version of the loop. So both tasks might get past the if_user_exists line, but one might add the user to the database first. So the second thread still thinks the user does not exist (since it didn't when it checked) but it does, and it might cause problems when the thread starts executing create_user etc

scheduler/state_machine/const.py
raar1: No, that sounds good! I just wondered about them angrily typing chat messages with no response :)

I think every 10 minutes might also be fine, but it's not really a huge problem

wbaccinelli:
@raar1 now the request check and acceptance is locket in case the previous run is not finished. But take a look at the solution

PR #93, opened 2023-06-06
357 query steps

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/357

### DIFF
--- a/Rasa_Bot/actions/definitions.py
+++ b/Rasa_Bot/actions/definitions.py
@@ -17,11 +17,13 @@ NICEDAY_API_ENDPOINT = os.getenv('NICEDAY_API_ENDPOINT')
 KEY_PATH = '/app/sensorprivatekey'
 
 if ENVIRONMENT == 'prod':
-    STEPS_URL = 'https://portal.rrdweb.nl/servlets/r2d2/v6.0.4/project/perfectfit/table/garmin_steps_day'
+    STEPS_URL = 'https://portal.rrdweb.nl/servlets/r2d2/v6.0.4/project/perfectfit/table' \
+                '/garmin_steps_day '
     # TODO: replace with actual url when this will be made available
     INTENSITY_URL = ''
 else:
-    STEPS_URL = 'https://www.rrdhost.nl/servlets/r2d2/v6.0.4/project/perfectfit/table/garmin_steps_day'
+    STEPS_URL = 'https://www.rrdhost.nl/servlets/r2d2/v6.0.4/project/perfectfit/table' \
+                '/garmin_steps_day '
     # TODO: replace with actual url when this will be made available
     INTENSITY_URL = ''
 
--- a/Rasa_Bot/actions/definitions.py
+++ b/Rasa_Bot/actions/definitions.py
@@ -6,9 +6,26 @@ from enum import Enum
 
 from dateutil import tz
 
+# dev or prod environment
+ENVIRONMENT = os.getenv('ENVIRONMENT')
+
 # load database url and niceday_api_endpoint variables
 DATABASE_URL = os.getenv('DATABASE_URL')
 NICEDAY_API_ENDPOINT = os.getenv('NICEDAY_API_ENDPOINT')
+
+# information for reading sensors data
+KEY_PATH = '/app/sensorprivatekey'
+
+if ENVIRONMENT == 'prod':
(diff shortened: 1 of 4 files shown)


Rasa_Bot/actions/definitions.py
```diff
@@ -6,10 +6,29 @@
...
+
 # load database url and niceday_api_endpoint variables
 DATABASE_URL = os.getenv('DATABASE_URL')
 NICEDAY_API_ENDPOINT = os.getenv('NICEDAY_API_ENDPOINT')
 
+# information for reading sensors data
+KEY_PATH = '/app/sensorprivatekey'
+
+if ENVIRONMENT == 'prod':
+    STEPS_URL = 'https://portal.rrdweb.nl/servlets/r2d2/v6.0.4/project/perfectfit/table' \
```
raar1, line 20: I'm wondering if this incredibly specific URL should be in the `.env` or here. Maybe it is fine here

Rasa_Bot/actions/helper.py
```diff
@@ -1255,4 +1262,72 @@ def get_weekly_intensity_minutes_goal_from_db(user_id: int) -> int:
...
+    """
+    Get the encoded JWT token for querying the sensors' data.
+    Args:
+        user_id: ID of the user whom data needs to be queried.
+
+    Returns: the encoded JWD token
+
+    """
+
+    with open(KEY_PATH, 'rb') as f:
```
raar1, line 1279: Maybe `SENSOR_KEY_PATH` could be more specific?

Rasa_Bot/actions/helper.py
```diff
@@ -1255,4 +1262,72 @@ def get_weekly_intensity_minutes_goal_from_db(user_id: int) -> int:
...
+        private_key = serialization.load_ssh_private_key(
+            f.read(), password=None, backend=default_backend()
+        )
+
+    encoded = jwt.encode({"sub": user_id, "iat": int(round(datetime.now().timestamp()))},
+                         private_key, algorithm="RS256")
```
raar1, line 1285: Does this assume that the private key from `KEY_PATH` is always RSA with SHA256? What happens if it is not?

Rasa_Bot/actions/helper.py
```diff
@@ -1255,4 +1262,72 @@ def get_weekly_intensity_minutes_goal_from_db(user_id: int) -> int:
...
+
+    return encoded
+
+
+# functions for sensors data querying
+def get_steps_data(user_id: int, start_date: date, end_date: date) -> List[Dict[Any, Any]]:
+    """
+    Get the encoded JWT token for querying the sensors' data.
```
raar1, line 1293: This seems to be a comment line from another function

Rasa_Bot/actions/helper.py
```diff
@@ -1255,4 +1262,72 @@ def get_weekly_intensity_minutes_goal_from_db(user_id: int) -> int:
...
+
+    token = get_jwt_token(user_id)
+
+    query_params = {'start': str(start_date),
+                    'end': str(end_date)}
+
+    headers = {TOKEN_HEADER: token}
+
+    res = requests.get(STEPS_URL, params=query_params, headers=headers, timeout=60)
+    res_json = res.json()
```
raar1, line 1311: Does this line always return json without erroring? I've worked with annoying systems before where certain errors on the remote caused the return of an html error page and then the `.json()` method would fail. If that's the case here, then a try-except might be needed

raar1 (approved):
Good work - nice to see the sensors stuff in there now! One or two minor comments below

Rasa_Bot/actions/helper.py
wbaccinelli: If not it will not work, but this is also was was required by RRD, right? Shall we consider different options?

Rasa_Bot/actions/helper.py
wbaccinelli: I should stop making large use of copy-paste XD

Rasa_Bot/actions/helper.py
wbaccinelli: The documentation of RRD states so. If I see an html page I can close the system for good ahahha. No, I will add the try-catch for robustness

PR #94, opened 2023-06-06
Oefening persuasion

nelealbers:
Fixes #https://github.com/PerfectFit-project/virtual-coach-issues/issues/356.

Based on feedback at the demo at the Persuasive Technology conference, replaced "activiteit" with "oefening" in persuasion in general activity dialog to be consistent with other dialogs.


wbaccinelli (approved):
Looks good!

PR #96, opened 2023-06-08
365 pause conversation

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/365
Fixes https://github.com/PerfectFit-project/testing-tickets/issues/11

The mechanism to pause and the conversations has been implemented. It makes use of the in-built `pause` and `resume` events of Rasa. The only solution to make the resume event work was using the post request to the rasa server, for which Celery has been adopted. Using the `PauseConversation()` and `ResumeConversation()` functions of the actions sdk was not working.


scheduler/celery_tasks.py
```diff
@@ -264,3 +265,79 @@ def trigger_intent(self,  # pylint: disable=unused-argument
...
+           user_id: int):
+    """
+    This task sends a resume event to Rasa.
+    Args:
+        user_id: the ID of the user to send the trigger to
+    """
+    endpoint = f'http://rasa_server:5005/conversations/{user_id}/tracker/events'
+    headers = {'Content-Type': 'application/json'}
+    data = '[{"event": "resume"}]'
+    requests.post(endpoint, headers=headers, data=data, timeout=60)
```
raar1, line 326: Can this fail/timeout if rasa_server is busy? (might happen with lots of users)

Would it just never resume?

raar1 (approved):
Give users some breathing time between dialogs :)

scheduler/celery_tasks.py
wbaccinelli: The users might be happy with this never resuming, but yes, I have to check this

PR #99, opened 2023-06-12
368 send inactive notification

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/368

### DIFF
--- a/scheduler/celery_tasks.py
+++ b/scheduler/celery_tasks.py
@@ -52,7 +52,7 @@ def setup_periodic_tasks(sender, **kwargs):  # pylint: disable=unused-argument
     # notify the FSM that a new day started
     sender.add_periodic_task(crontab(hour=00, minute=00), notify_new_day.s(datetime.today()))
     # check if the user is active and send notification
-    sender.add_periodic_task(crontab(hour=15, minute=40), check_inactivity.s())
+    sender.add_periodic_task(crontab(hour=10, minute=00), check_inactivity.s())
     # check if a dialog has been completed
     sender.add_periodic_task(MAXIMUM_DIALOG_DURATION, check_dialogs_status.s())
     # check if new connections are pending and, in case, accept them
@@ -285,7 +285,11 @@ def trigger_intent(self,  # pylint: disable=unused-argument
     headers = {'Content-Type': 'application/json'}
     params = {'output_channel': 'niceday_trigger_input_channel'}
     data = '{"name": "' + trigger + '" }'
-    requests.post(endpoint, headers=headers, params=params, data=data, timeout=60)
+    response = requests.post(endpoint, headers=headers, params=params, data=data, timeout=60)
+
+    if response.status_code != 200:
+        # if the request failed resend it
+        requests.post(endpoint, headers=headers, params=params, data=data, timeout=60)
 
     if dialog_status is not None:
         # the state machine status as to be marked as not running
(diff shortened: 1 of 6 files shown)


raar1:
Looks like there's some conflicts - could you quickly fix those then I can review?

scheduler/celery_tasks.py
```diff
@@ -245,22 +266,28 @@ def user_trigger_dialog(self,  # pylint: disable=unused-argument
...
+    user_fsm = get_user_fsm(user_id)
+    current_dialog_state = get_dialog_state(user_fsm)
+    if current_dialog_state == RUNNING:
+        return
+
     endpoint = f'http://rasa_server:5005/conversations/{user_id}/trigger_intent'
     headers = {'Content-Type': 'application/json'}
     params = {'output_channel': 'niceday_trigger_input_channel'}
     data = '{"name": "' + trigger + '" }'
     requests.post(endpoint, headers=headers, params=params, data=data, timeout=60)
```
raar1, line 288: If this times out, does something bad happen? Or will it just end up being sent again later?

scheduler/celery_utils.py
```diff
@@ -31,6 +31,39 @@ def check_if_user_exists(user_id: int) -> bool:
...
+        .filter(
+            UserInterventionState.users_nicedayuid == user_id,
+            UserInterventionState.completed.is_(True),
+            UserInterventionState.last_time > latest_date
+        )
+        .first()
+    )
+
+    # there is at least one completed dialog in past days
+    if last_completed is not None:
```
raar1, line 61: I suppose this is to avoid bothering the user when they have just signed up? Or for something else?

raar1 (approved):
Looks good!

scheduler/celery_tasks.py
```diff
@@ -48,8 +49,13 @@ def setup_periodic_tasks(sender, **kwargs):  # pylint: disable=unused-argument
     When celery is ready, the watchdogs for the new day notification
     and for the dialogs status check are started
     """
+    # notify the FSM that a new day started
     sender.add_periodic_task(crontab(hour=00, minute=00), notify_new_day.s(datetime.today()))
+    # check if the user is active and send notification
+    sender.add_periodic_task(crontab(hour=15, minute=40), check_inactivity.s())
```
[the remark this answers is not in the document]
wbaccinelli: Because I am stupid and I did not change this after testing XD

scheduler/celery_tasks.py
```diff
@@ -111,6 +117,21 @@ def check_dialogs_status(self):  # pylint: disable=unused-argument
...
+@app.task
+def check_inactivity():
+    """
+    This task checks for all the users if they have been inactive for at least 10 days.
+    If they have been inactive, the correspondent notification is sent
+    Args:
+    """
+    current_date = date.today()
+    state_machines = get_all_fsm()
+    for item in state_machines:
```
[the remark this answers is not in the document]
wbaccinelli: Yes, this is a good point. Also, the users that are finished with the trial should be removed. at the moment, this doesn't happen.

scheduler/celery_tasks.py
wbaccinelli: If the post fails, is might be a problem. I am adding a check

scheduler/celery_utils.py
wbaccinelli: No, but I didn't think about this. I have to refine this to avoid people to get this notification immediately

wbaccinelli: But actually, this will be None if they never completed a dialog, so should be fine

PR #102, opened 2023-06-14
adjusted speed to message length

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/360 

The messages sending thorough the API happens when the dialogs are externally triggered. The response is sent to the API through the `custom_channels.py`. A delay between one message and the other has been added there.

### DIFF
--- a/Rasa_Bot/custom_channels.py
+++ b/Rasa_Bot/custom_channels.py
@@ -17,7 +17,7 @@ NICEDAY_API_URL = os.getenv('NICEDAY_API_ENDPOINT')
 # proportional factor between the number of words in a message and the time to wait before the
 # next message is delivered
 WORDS_PER_SECOND = 5
-# maximum delay between a message and the next one
+# maximum delay, in seconds, between a message and the next one
 MAX_DELAY = 10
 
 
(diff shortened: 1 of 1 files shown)


Rasa_Bot/custom_channels.py
```diff
@@ -13,6 +14,11 @@
 
 
 NICEDAY_API_URL = os.getenv('NICEDAY_API_ENDPOINT')
+# proportional factor between the number of words in a message and the time to wait before the
+# newt message is delivered
```
raar1, line 18:
```suggestion
# next message is delivered
```

Rasa_Bot/custom_channels.py
```diff
@@ -13,6 +14,11 @@
...
+WORDS_PER_SECOND = 5
```
raar1, line 19: Haha, ok, so this is the fix for the PF app spamming users with 6000 messages a minute like a deranged stalker?

Rasa_Bot/custom_channels.py
```diff
@@ -13,6 +14,11 @@
...
+# maximum delay between a message and the next one
+MAX_DELAY = 10
```
raar1, line 21: Also units of seconds, I assume?

raar1 (approved):

Rasa_Bot/custom_channels.py
wbaccinelli: ahahah yes, now it should be less annoying (if possible).

Rasa_Bot/custom_channels.py
wbaccinelli: yep, I'm updating the definition

PR #110, opened 2023-06-28
354 disconnect user

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/354

Makes use of https://github.com/PerfectFit-project/niceday-components/pull/18 and https://github.com/PerfectFit-project/niceday_client/pull/10

Added the action to disconnect the user from the therapist at the end of the intervention. 
As the chat might still work during the active session, even if the user has been disconnected, a new state in the FSM has been added, where the dialogs are not triggered. The state is activated after the closing dialog ends.


Rasa_Bot/actions/actions_closing_dialog.py
```diff
@@ -271,6 +272,21 @@ async def run(self, dispatcher, tracker, domain):
...
+    """Disconnect user from therapist."""
+
+    def name(self):
+        return "action_disconnect_user"
+
+    async def run(self, dispatcher, tracker, domain):
+        user_id = tracker.current_state()['sender_id']
+
+        client = NicedayClient(NICEDAY_API_ENDPOINT)
+        client.remove_contact(user_id)
```
raar1, line 285: Can this fail? If so does anything retry? Will something bad happen if a user stays connected anyway?

wbaccinelli: At this point, everything can fail in any moment, I'm afraid. Incase the use will not be disconnected, he/she will be able to keep on spamming coach Sam, but Sam will simply not respond.

raar1: OK that doesn't seem too serious then

PR #115, opened 2023-06-30
353 restore tasks

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/353

Note: when the scheduler service is restarted, if the tasks scheduled are still stored in redis, they will be automatically queued again. This takes few seconds after the worker is ready, but there are no signals dispatched by celery to detect this event. As a workaround, the check is run 1 minute after the worker is ready. Not very clean, but I found no better alternatives.


scheduler/celery_utils.py
```diff
@@ -318,6 +318,28 @@ def get_user_fsm(user_id: int) -> StateMachine:
     return user_fsm
 
 
+def get_scheduled_task_from_db() -> List[UserInterventionState]:
+    """
+    Get the list of all the tasks scheduled and not yet executed as they are stored in the DB.
```
raar1, line 323: So is there a reason why a task that was scheduled and has now been executed would not simply be removed from the db upon completion?

raar1 (approved):
Seems good to me

scheduler/celery_utils.py
wbaccinelli: I think that there might be cases in which this could happen. For example, if there is a failure during the execution of the task. Or if there are bugs (might well be)

PR #117, opened 2023-07-04
Add function to calculate intensity minutes

bscheltinga:
- Fixes a bug in using API to load steps data for specific dates
- Added function to get intensity minutes from Senseeact DB


wbaccinelli (approved):
Nice work! It might be good to investigate the possibility to filter the results in the query already. I see that in [the documentation](https://www.rrdhost.nl/senseeact/docs/api/controllers/project-read.html#project-table-table-filter-get) is mentioned already

bscheltinga:
> Nice work! It might be good to investigate the possibility to filter the results in the query already. I see that in [the documentation](https://www.rrdhost.nl/senseeact/docs/api/controllers/project-read.html#project-table-table-filter-get) this mentioned already

Both myself and @wbaccinelli took a look at the filter query. But we won't get the data we wanted using this query

PR #123, opened 2023-07-14
fixing everything

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/382

### DIFF
--- a/scheduler/celery_tasks.py
+++ b/scheduler/celery_tasks.py
@@ -11,7 +11,7 @@ from django.core.cache import cache
 from state_machine.state_machine import EventEnum, Event
 from state_machine.const import (REDIS_URL, TIMEZONE, MAXIMUM_DIALOG_DURATION, NICEDAY_API_ENDPOINT,
                                  RUNNING, EXPIRED, NOTIFY, INVITES_CHECK_INTERVAL,
-                                 MAXIMUM_INACTIVE_DAYS, ENVIRONMENT, WORDS_PER_SECOND, MAX_DELAY)
+                                 MAXIMUM_INACTIVE_DAYS, WORDS_PER_SECOND, MAX_DELAY)
 from typing import Optional
 from celery_utils import (check_if_task_executed, check_if_user_active, check_if_user_exists,
                           create_new_user, get_component_name, get_user_fsm, get_dialog_state,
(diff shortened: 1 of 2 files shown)


scheduler/celery_tasks.py
```diff
@@ -395,11 +393,38 @@ def resume_and_trigger(self,  # pylint: disable=unused-argument
     if response_resume.status_code != 200:
         raise Exception()
 
+    response_intent = send_trigger(user_id, trigger)
+    if response_intent != 200:
+        print('Exception, retrying')
```
raar1, line 398: Can we log this instead of printing it, or is that not appropriate here? Also I would make the message slightly more descriptive - maybe "Exception during resume_and_trigger" or something, to at least say which function it was happening it (same for the others).

scheduler/celery_tasks.py
```diff
@@ -395,11 +393,38 @@ def resume_and_trigger(self,  # pylint: disable=unused-argument
...
 
-    if response_intent.status_code != 200:
-        raise Exception()
+    res_json = response_intent.json()
+
+    for mes in res_json['messages']:
+        recipient_id = mes['recipient_id']
+        message = mes['text']
+        client.post_message(int(recipient_id), message)
+        if ENVIRONMENT == 'prod':
```
raar1, line 425: There are no delays simulated in testing?

scheduler/celery_tasks.py
```diff
@@ -395,11 +393,38 @@ def resume_and_trigger(self,  # pylint: disable=unused-argument
...
 
+    response_intent = send_trigger(user_id, trigger)
+    if response_intent != 200:
+        print('Exception, retrying')
+        raise Exception()
+
+
+def send_trigger(user_id: int, trigger: str):
+    """
+    Prepare and sent the HTTP post request to rasa for triggering an intent.
```
raar1, line 404:
```suggestion
    Prepare and send the HTTP post request to rasa for triggering an intent.
```

raar1 (approved):
OK, nice. So, am I understanding correctly that you have moved the message sending code (that used to be in Rasa) to a celery task, so it's now fully async? The delay issue is no longer noticeable?

wbaccinelli:
> OK, nice. So, am I understanding correctly that you have moved the message sending code (that used to be in Rasa) to a celery task, so it's now fully async? The delay issue is no longer noticeable?

@raar1 Exactly, and it runs so smoothly now that I could cry.

scheduler/celery_tasks.py
wbaccinelli: Absolutely yes, I will do this.

scheduler/celery_tasks.py
wbaccinelli: Yeah, maybe I'd better keep it always active indeed.

PR #124, opened 2023-07-14
Enhance plot of figure with steps and step goals

bscheltinga:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/384

Note that it can still happen that the text for the actual steps and step goal can overlap, in case the number of steps and goal is around 3000... Check the before and after image in the issue: https://github.com/PerfectFit-project/virtual-coach-issues/issues/384


wbaccinelli (approved):
Nice, the image is looks better

PR #125, opened 2023-07-14
333 configure step count in dialogs

bscheltinga:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/333


bscheltinga:
I spotted: https://github.com/PerfectFit-project/virtual-coach-rasa/blob/a1879481a5f7a486d86a709853b16a5eaeb3df61/Rasa_Bot/actions/actions_weekly_reflection.py#L201

I do not understand why this is here...

bscheltinga:
I realize that most functions I created will not work if no data is returned from the API. Also, I tested all functions in isolation, as spinning up the system would not work as there is no data.

Rasa_Bot/actions/actions_weekly_reflection.py
```diff
@@ -287,13 +311,14 @@ def name(self):
         return "action_show_pa_overview"
 
     async def run(self, dispatcher, tracker, domain):
-        ## TODO retrieve step goal per day and dates, placeholder for now
-
+        # Get data from API
+        user_id = int(tracker.current_state()['sender_id'])
+        end = datetime.datetime.now()
+        start = end - datetime.timedelta(days=16)
```
wbaccinelli, line 317: same here

Rasa_Bot/actions/helper.py
```diff
@@ -1226,23 +1226,65 @@ def get_faik_text(user_id):
...
+    end = datetime.now()
+    start = end - timedelta(days=9)
+    steps_data = get_steps_data(user_id=user_id, start_date=start, end_date=end)
+
+    for day in steps_data:
+        steps_per_day.append(day['steps'])
+
+    if len(steps_per_day) < 9:
+        while len(steps_per_day) < 9:
+            steps_per_day.append(np.mean(steps_per_day))  # Add with the avg value up to 9 values
```
wbaccinelli, line 1257:
```suggestion
    for _ in range(9 - len(steps_per_day)):
        steps_per_day.append(np.mean(steps_per_day))  # Add with the avg value up to 9 values
```

It's always better to avoid the while cycles.
Also, the mean will be updated every time you update the list, is that fine?

Rasa_Bot/actions/helper.py
```diff
@@ -1226,23 +1226,65 @@ def get_faik_text(user_id):
...
+        step_goal (int, float, list): The input value(s) to be transformed. It can be a single
+                                      value or a list of values.
+
+    Returns:
+        int, float, list: If `pa_goal` is a single value, the function returns the transformed
+                          value according to the minimum and maximum bounds. If `pa_goal` is a
+                          list of values, the function returns a list of transformed values
+    """
+    min_val = 2000
+    max_val = 10000
```
wbaccinelli, line 1281: Also this values should be moved to the constants

Rasa_Bot/actions/helper.py
```diff
@@ -1373,3 +1421,67 @@ def get_intensity_minutes_data(user_id: int,
...
+
+    # Resample the data to a day-to-day base. Add nans for empty dates
+    df = df.asfreq('D')
+
+    # Calculate the step goals
+    steps = list(df.steps)
+    step_goals = []
+    for i in range(7):
+        steps_nine_days = steps[i:i + 9]
+        steps_nine_days = [x for x in steps_nine_days if not pd.isna(x)]  # Remove NaN values
```
wbaccinelli, line 1467:
```suggestion
    steps_nine_days = [steps[i:i + 9] for i in range(7) if not pd.isna(steps[i:i + 9])]  # Remove NaN values
```

Not tested, but I guess that this does the same thing

wbaccinelli (approved):
Well done! I left some suggestions. Probably other things can be optimized, but this will work anyway

Rasa_Bot/actions/helper.py
bscheltinga: I commited the suggestion. The mean will not change if the added sample is equal to the mean.

Rasa_Bot/actions/helper.py
bscheltinga: I think this suggestion will not work. The for loop includes `step_goals.append(int(round(steps_nine_days[5], -1)))  # Definition of the goal`

wbaccinelli (approved):
The changes look good to me.

sensor_api/sensorapi/connector.py
```diff
@@ -2,11 +2,13 @@
 import logging
 import os
 import requests
+import pandas as pd
+import numpy as np
```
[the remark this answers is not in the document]
bscheltinga: Good catch :)

PR #128, opened 2023-07-17
check if effor slot is not none

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/374

### DIFF
--- a/Rasa_Bot/actions/actions_general_activity.py
+++ b/Rasa_Bot/actions/actions_general_activity.py
@@ -637,6 +637,7 @@ class SavePersuasionToDatabase(Action):
         store_dialog_closed_answer_to_db(user_id, answer_value=message_idx + 2,
                                          question_id=question_id)
         # +1 since the lowest value is 0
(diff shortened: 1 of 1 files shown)


Rasa_Bot/actions/actions_general_activity.py
```diff
@@ -637,8 +637,10 @@ async def run(self, dispatcher, tracker, domain):
         store_dialog_closed_answer_to_db(user_id, answer_value=message_idx + 2,
                                          question_id=question_id)
         # +1 since the lowest value is 0
-        store_dialog_closed_answer_to_db(user_id, answer_value=effort + 1,
-                                         question_id=DialogQuestionsEnum.PERSUASION_EFFORT.value)
+        if effort is not None:
```
nelealbers, line 640:
```suggestion
        # Effort is None the first time the general activity dialog is run for a user
        if effort is not None:
```

nelealbers (approved):
Looks good! Just added a comment so we remember when the slot might be None.

PR #131, opened 2023-07-20
changed celery task in the actions

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/388
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/391
Fixes: https://github.com/PerfectFit-project/testing-tickets/issues/22
Fixes: https://github.com/PerfectFit-project/testing-tickets/issues/23

The problem was that the type of task that was sent to the scheduler by the video dialogs was wrong. What happened was that the just the generic `watch_video_dialog` and `reschedule_preparation_phase` dialogs where stored in the FSM. If the dialog expired, and the FSM rescheduled it, the wrong element was saved in the DB, and it was never marked as completed.

### DIFF
--- a/Rasa_Bot/actions/actions_preparation_phase.py
+++ b/Rasa_Bot/actions/actions_preparation_phase.py
@@ -9,7 +9,7 @@ from virtual_coach_db.helper.definitions import (Components,
                                                  DialogExpectedDuration)
 
 from . import validator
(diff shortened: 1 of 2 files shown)


nelealbers (approved):
Looks good!

PR #132, opened 2023-07-20
390 pause after profile creation

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/390
Fixes: https://github.com/PerfectFit-project/testing-tickets/issues/19

After the completion of the profile creation dialog, the conversation is now paused for 30 seconds before triggering the next dialog. The pause is necessary for two reasons: 
- we cannot triggered immediately the next dialog, because the messages of the profile will be dispatched also during the new dialog (due to the time delay used between the messages)
- if the user types something during between the two dialogs, mixed messages will be received


nelealbers (approved):
Looks good - just added a comment to help people like me understand what it means to "acknowledge" the FSM.

PR #136, opened 2023-07-27
403 remove hrs preparation

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/403
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/400

Changed the menu selection. Now the options are split in separate utterances, and the ones that are not always available are checked before showing them. As a safety layer, this is first done in the rasa actions and then in the state machine.

The relapse dialog is no more available before the execution phase.

The order of the menu utterance has been modified, and the default first utterance has been adapted.

### DIFF
--- a/scheduler/state_machine/controller.py
+++ b/scheduler/state_machine/controller.py
@@ -250,12 +250,12 @@ class GoalsSettingState(State):
                           phase=1)
 
     def on_user_trigger(self, dialog):
-        # in this phase a dialog can be continued
-        if dialog == Components.RELAPSE_DIALOG:
-            run_option_menu(self.user_id)
         # the relapse dialog is not available in this phase
-        elif dialog == Components.CONTINUE_UNCOMPLETED_DIALOG:
+        if dialog == Components.CONTINUE_UNCOMPLETED_DIALOG:
             run_uncompleted_dialog(self.user_id)
+        # in this phase a dialog can be continued
+        elif dialog == Components.RELAPSE_DIALOG:
+            run_option_menu(self.user_id)
         else:
             plan_and_store(user_id=self.user_id,
                            dialog=dialog,
@@ -365,8 +365,8 @@ class BufferState(State):
         if dialog == Components.CONTINUE_UNCOMPLETED_DIALOG:
             run_uncompleted_dialog(self.user_id)
         # the relapse dialog is not available in this phase
-        elif dialog == Components.CONTINUE_UNCOMPLETED_DIALOG:
-            run_uncompleted_dialog(self.user_id)
+        elif dialog == Components.RELAPSE_DIALOG:
+            run_option_menu(self.user_id)
         else:
             plan_and_store(user_id=self.user_id,
                            dialog=dialog,


Rasa_Bot/domain/domain_idle_commands.yml
```diff
@@ -15,7 +15,7 @@ responses:
   # For low next action confidence, "utter_default" is automatically called.
   # For low NLU confidence, we have defined a rule.
   utter_default:
-    - text: "Sorry, ik begrijp (nog) niet wat je bedoelt, kun je me het op een andere manier duidelijk maken?"
+    - text: "Goed om van je te horen! Er zijn verschillende dingen die je kunt doen."
```
nelealbers, line 18:
```suggestion
    - text: "Goed om van je te horen! Er zijn verschillende dingen die je kunt doen:"
```

scheduler/state_machine/controller.py
```diff
@@ -365,6 +364,9 @@ def on_new_day(self, current_date: date):
     def on_user_trigger(self, dialog: str):
         if dialog == Components.CONTINUE_UNCOMPLETED_DIALOG:
             run_uncompleted_dialog(self.user_id)
+        # the relapse dialog is not available in this phase
+        elif dialog == Components.CONTINUE_UNCOMPLETED_DIALOG:
```
nelealbers, line 368: Why is this if-statement here twice? I.e., why are we checking for dialog == Components.CONTINUE_UNCOMPLETED_DIALOG twice here? Based on the comment above maybe the second statement should be something about the relapse dialog instead?

scheduler/state_machine/controller.py
```diff
@@ -255,7 +251,10 @@ def on_dialog_rescheduled(self, dialog, new_date):
 
     def on_user_trigger(self, dialog):
         # in this phase a dialog can be continued
-        if dialog == Components.CONTINUE_UNCOMPLETED_DIALOG:
+        if dialog == Components.RELAPSE_DIALOG:
+            run_option_menu(self.user_id)
+        # the relapse dialog is not available in this phase
```
nelealbers, line 256: This comment does not seem to match what is happening below it (i.e., continuing an uncompleted dialog).

scheduler/state_machine/controller.py
```diff
@@ -255,7 +251,10 @@ def on_dialog_rescheduled(self, dialog, new_date):
 
     def on_user_trigger(self, dialog):
         # in this phase a dialog can be continued
```
nelealbers, line 253: This comment also does not seem to match. I think the 2 comments might have to be swapped?

scheduler/state_machine/controller.py
wbaccinelli: I pasted the wrong line... thank you for spotting this!

scheduler/state_machine/controller.py
wbaccinelli: Yes, correct!

PR #139, opened 2023-08-01
update db version

nelealbers:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/409.

Update db version after updating resource texts in db.


wbaccinelli (approved):
Alright

PR #140, opened 2023-08-01
396 hrs autocomplete on expire

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/396
Fixes: https://github.com/PerfectFit-project/testing-tickets/issues/25
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/393
Fixes: https://github.com/PerfectFit-project/testing-tickets/issues/26

- Storing the branch chosen by the user in the relapse dialog as a new closed answer
- Moved the expiration case management from the celery_tsks to the single state for clarity
- Improved intents management in ehbo 
- In the relapse state, the dialog is auto-completed in case a user-initiated relapse dialog expires, out of the relapse branch

### DIFF
--- a/scheduler/state_machine/controller.py
+++ b/scheduler/state_machine/controller.py
@@ -25,7 +25,7 @@ celery = Celery(broker=REDIS_URL)
 class OnboardingState(State):
 
     def __init__(self, user_id: int):
-        super().__init__(user_id)
+        super().__init__(user_id, celery)
         self.state = State.ONBOARDING
         self.user_id = user_id
         self.new_state = None
@@ -150,7 +150,7 @@ class OnboardingState(State):
 class TrackingState(State):
 
     def __init__(self, user_id):
-        super().__init__(user_id)
+        super().__init__(user_id, celery)
         self.user_id = user_id
         self.state = State.TRACKING
 
@@ -217,7 +217,7 @@ class TrackingState(State):
 class GoalsSettingState(State):
 
     def __init__(self, user_id):
-        super().__init__(user_id)
+        super().__init__(user_id, celery)
         self.user_id = user_id
         self.state = State.GOALS_SETTING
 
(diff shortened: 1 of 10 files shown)


scheduler/state_machine/state_machine_utils.py
```diff
@@ -279,6 +280,75 @@ def get_intervention_component(intervention_component_name: str) -> Intervention
...
+            DialogQuestions.question_id == DialogQuestionsEnum.RELAPSE_SMOKE_HRS_LAPSE_RELAPSE.value
+        )
+        .order_by(DialogClosedAnswers.datetime.desc())
+        .limit(1)
+        .one_or_none()
+    )
+
+    # if there are no answers or the answer came before the last execution of the dialog,
+    # return None
+    if last_answer is None or last_answer.datetime < last_execution.last_time:
```
nelealbers, line 330: So the last_time value is always only set when a dialog is completed, not yet when it is started?

scheduler/state_machine/controller.py
```diff
@@ -594,6 +604,31 @@ def on_dialog_completed(self, dialog):
...
+                                   phase_id=3)
+
+            # let the fms know that the dialog is considered as not running anymore
+            update_fsm_dialog_running_status(self.user_id, False)
+            # go back to the execution
+            logging.info('Relapse completed, back to execution')
+            self.set_new_state(ExecutionRunState(self.user_id))
+
+        else:
+            next_day = datetime.now() + timedelta(days=1)
```
nelealbers, line 625: I was wondering if adding one day is a good idea, since there might be a structural reason why people did not complete the dialog now. For example, could we just use people's preferred time from the profile creation dialog for this? But I suppose especially at the start people might not yet have such a preferred time. And maybe this is just overcomplicating things.

nelealbers (approved):
Looks good, just added two thoughts/comments.

scheduler/state_machine/state_machine_utils.py
wbaccinelli: `last_time ` is updated when the dialog is started, this is why if the answer came before that, it is for sure from another run

nelealbers: Ah I seee, that makes sense!

PR #141, opened 2023-08-02
Avoid messages in the night

wbaccinelli:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/392
Fixes: https://github.com/PerfectFit-project/testing-tickets/issues/24

- Check for the time before scheduling and before triggering a scheduled task. If the task occurs in nighttime, it is moved to the next morning 
- Avoid checking for the dialog status in the night.

### DIFF
--- a/scheduler/celery_tasks.py
+++ b/scheduler/celery_tasks.py
@@ -163,7 +163,7 @@ def check_dialogs_status(self):  # pylint: disable=unused-argument
     # this check should not run between 23 and 7
     current_date = datetime.now(tz=TIMEZONE)
 
-    if current_date.hour == 23 or current_date.hour < MORNING_TIME:
+    if 0 < current_date.hour < MORNING_TIME:
         return
 
     logging.info("Checking the dialogs status")
--- a/scheduler/celery_tasks.py
+++ b/scheduler/celery_tasks.py
@@ -159,6 +159,13 @@ def check_dialogs_status(self):  # pylint: disable=unused-argument
     This task verifies if there are uncompleted dialogs and, in case, reschedules them.
     The task is rescheduled every maximum duration of the dialog time
     """
+
+    # this check should not run between 23 and 7
+    current_date = datetime.now(tz=TIMEZONE)
+
+    if current_date.hour == 23 or current_date.hour < 7:
+        return
+
     logging.info("Checking the dialogs status")
 
     state_machines = get_all_fsm()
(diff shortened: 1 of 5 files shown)


nelealbers:
@wbaccinelli When we offer people the opportunity to reschedule dialogs, we send them at 7am if they choose the morning (https://github.com/PerfectFit-project/virtual-coach-rasa/blob/main/Rasa_Bot/actions/definitions.py). For consistency, it might be good to also use 7am here as the "morning time".

wbaccinelli:
> @wbaccinelli When we offer people the opportunity to reschedule dialogs, we send them at 7am if they choose the morning (https://github.com/PerfectFit-project/virtual-coach-rasa/blob/main/Rasa_Bot/actions/definitions.py). For consistency, it might be good to also use 7am here as the "morning time".

@nelealbers You're absolutely right! Will do that

scheduler/celery_tasks.py
```diff
@@ -243,9 +250,15 @@ def reschedule_dialog(user_id: int, intervention_component_name: str, new_date:
     """
 
     logging.info('Celery received a dialog rescheduling')
+
+    # check if the scheduled time is in the night (i.e., after midnight and before 6)
+    # In case it is, reschedule for the morning.
+    if 0 <= new_date.hour <= MORNING_TIME:
```
bscheltinga, line 256: To be consistent with line 166, shouldn't this be between ==23 or  <MORNING_TIME?

scheduler/celery_tasks.py
```diff
@@ -305,6 +318,16 @@ def trigger_scheduled_intervention_component(self,
     # retrieve the name of the component
     name = get_component_name(trigger)
 
+    # check if the current time is in the night (i.e., after midnight and before 6)
+    # In case it is, reschedule for the morning.
+    current_date = datetime.now(tz=TIMEZONE)
+    if 0 <= current_date.hour <= 7:
```
bscheltinga, line 324: Same here: ==23 and < MORNING_Time

scheduler/state_machine/const.py
```diff
@@ -48,3 +48,6 @@
 RUNNING = 1
 EXPIRED = 2
 NOTIFY = 3
+
+# definitions of morning hours
+MORNING_TIME = 7
```
bscheltinga, line 53: In actions_rescheduling_dialog.py, MORNING_TIME is defined as 8, should we be consistent here?

https://github.com/PerfectFit-project/virtual-coach-rasa/blob/02d519eb7db70e809d4d03ca227e4b5bf11eba7b/Rasa_Bot/actions/actions_rescheduling_dialog.py#L83

bscheltinga (approved):
Hi Walter, I spotted some inconsistenties, but I'm not sure whether these are necessarily bad.

scheduler/celery_tasks.py
wbaccinelli: Right, this doesn't have a lot of sense

scheduler/state_machine/const.py
wbaccinelli: Well spotted, we had the definitions but they were not used. I am adapting also the actions to make use of those.

PR #144, opened 2023-08-03
408 exit relapse dialog

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/408
Fixes https://github.com/PerfectFit-project/testing-tickets/issues/53

- Moved possibility to leave the dialog in the first question.
- Added `current_intervention_component` slot setting for correctly marking the dialog completion.
- Added a final sentence in case the user leaves the dialog in the first question.

### DIFF
--- a/Rasa_Bot/data/rules/rules_relapse_dialog.yml
+++ b/Rasa_Bot/data/rules/rules_relapse_dialog.yml
@@ -8,6 +8,7 @@ rules:
 - rule: Trigger relapse dialog
   steps:
   - intent: EXTERNAL_relapse_dialog
+  - action: action_set_slot_relapse_dialog
   - action: utter_relapse_intro_codeword
   - action: utter_relapse_intro_offer_help_pt1
   - action: utter_relapse_intro_offer_help_pt2
--- a/Rasa_Bot/data/rules/rules_relapse_dialog.yml
+++ b/Rasa_Bot/data/rules/rules_relapse_dialog.yml
@@ -16,6 +16,18 @@ rules:
   - active_loop: smoke_or_pa_form
   wait_for_user_input: false
 
+## rigger relapse dialog  - exit
+- rule: Relapse phase triggered smoke
+  condition:
+  - active_loop: smoke_or_pa_form
+  steps:
+  - action: smoke_or_pa_form
+  - active_loop: null
+  - slot_was_set:
+    - smoke_or_pa: 3
+  - action: mark_dialog_as_completed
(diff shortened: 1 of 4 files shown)


Rasa_Bot/data/rules/rules_relapse_dialog.yml
```diff
@@ -8,6 +8,7 @@ rules:
 - rule: Trigger relapse dialog
   steps:
   - intent: EXTERNAL_relapse_dialog
+  - action: action_set_slot_relapse_dialog
```
bscheltinga, line 11: Is this needed? Do we need to store if people had this conversation accidentally? Also, do you mean this action:
https://github.com/PerfectFit-project/virtual-coach-rasa/blob/09878cff88aa542033febb94e5f0fed42902bfdc/Rasa_Bot/actions/actions_relapse_dialogs.py#L128-L135

I think the class name and string in the return are mismatching, I found that a bit confusing

Rasa_Bot/domain/domain_relapse_dialogs.yml
```diff
@@ -1060,6 +1065,7 @@ responses:
 actions:
   - action_check_reasons
   - action_launch_goal_setting
+  - action_set_slot_relapse_dialog
```
bscheltinga, line 1068: This can be removed if it is unused

bscheltinga (approved):
Nice looks good :)

Rasa_Bot/data/rules/rules_relapse_dialog.yml
wbaccinelli: It is needed because when the dialog is started we write this in the DB, so we also need to know if this is over. For this we need the slot setting, for using the `mark_dialog_as_completed` action.
I refer to [this action](https://github.com/PerfectFit-project/virtual-coach-rasa/blob/09878cff88aa542033febb94e5f0fed42902bfdc/Rasa_Bot/actions/actions_execution_dialogs.py#L47) which should be placed in a different file actually. The one you mention is for the hrs branch

bscheltinga: Ok that makes sense. That `mark_dialog_as_completed` is in another file is probably why I could not find it.

wbaccinelli: Yes, that is in actions_common, because we use it for all the dialogs.

PR #145, opened 2023-08-03
merged first two utterances

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/397

The second message was delayed (yes, the usual imposed delay...) and the following action, which was the triggering of the video dialog was too fast. Merging the first and the second message is the fastest and simplest solution. More complicated things can be done, but I would definitely not go that way.


bscheltinga (approved):
Awesome!

wbaccinelli:
@bscheltinga I was inspired by you :)

PR #152, opened 2023-08-07
start the goal setting state if the tracking time is over

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/417
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/412
Fixes https://github.com/PerfectFit-project/testing-tickets/issues/44

- The goal setting dialog is triggered on the expected date, even if the short version of the future self has not been completed.
- If the date for showing the short version of the future self video has expired, it is not shown. As this is checked when the long version has been completed, we avoid a repetition of the same content.

### DIFF
--- a/scheduler/state_machine/controller.py
+++ b/scheduler/state_machine/controller.py
@@ -124,9 +124,10 @@ class OnboardingState(State):
         fs_time = create_new_date(start_date=start_date,
                                   time_delta=FUTURE_SELF_INTRO)
 
-        # if the expected scheduling time is already passed, launch the component now
+        # if the expected scheduling time is already passed, don't show
+        # (the long version has just been shown)
         if fs_time.date() <= date.today():
-            fs_time = None
+            return
 
         plan_and_store(user_id=self.user_id,
                        dialog=Components.FUTURE_SELF_SHORT,
--- a/scheduler/state_machine/controller.py
+++ b/scheduler/state_machine/controller.py
@@ -200,6 +200,8 @@ class TrackingState(State):
 
     def on_new_day(self, current_date: date):
         logging.info('current date: %s', current_date)
+        self.check_if_end_date(current_date)
+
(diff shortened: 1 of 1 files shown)


scheduler/state_machine/controller.py
```diff
@@ -124,9 +124,10 @@ def schedule_next_dialogs(self):
         fs_time = create_new_date(start_date=start_date,
                                   time_delta=FUTURE_SELF_INTRO)
 
-        # if the expected scheduling time is already passed, launch the component now
+        # if the expected scheduling time is already passed, don't show
```
nelealbers, line 127:
```suggestion
        # if the expected scheduling time is already passed, don't show the short version of the future self dialog

```

scheduler/state_machine/controller.py
```diff
@@ -124,9 +124,10 @@ def schedule_next_dialogs(self):
         fs_time = create_new_date(start_date=start_date,
                                   time_delta=FUTURE_SELF_INTRO)
 
-        # if the expected scheduling time is already passed, launch the component now
+        # if the expected scheduling time is already passed, don't show
+        # (the long version has just been shown)
         if fs_time.date() <= date.today():
```
nelealbers, line 129: I do not yet quite understand this. So schedule_next_dialogs() is only called after the long version of the future self dialog has been shown? Just wondering because it has such a generic name.

nelealbers (approved):
Looks good. I just do not yet quite understand how it works that the short future self version is skipped when the long one has just been shown. I suppose the corresponding function schedule_next_dialogs() is only called after the long future self version?

scheduler/state_machine/controller.py
wbaccinelli: Exactly, this is called on the completion of the long version of the future self video (line 89)

PR #153, opened 2023-08-08
422 correction weekly reflection

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/422

- Corrected the validation of `how_feel_smoke` and `hrs_feeling_slot` in relapse dialog.
- Added the graph visualization in missing points.
- Updated dependencies to the latest DB version

### DIFF
--- a/Rasa_Bot/actions/actions_relapse_dialogs.py
+++ b/Rasa_Bot/actions/actions_relapse_dialogs.py
@@ -897,11 +897,11 @@ class ValidateWhatDoingHowFeelSmokeForm(FormValidationAction):
         if last_utterance != 'utter_ask_how_feel_smoke':
             return {"how_feel_smoke": None}
 
-        valid = validator.validate_list(value, 0, 9)
+        valid = validator.validate_list(value, 1, 9)
 
         if not valid:
             dispatcher.utter_message(response="utter_did_not_understand")
-            dispatcher.utter_message(response="utter_please_answer_numbers")
+            dispatcher.utter_message(response="utter_please_answer_1_to_9")
             return {"how_feel_smoke": None}
 
         return {"how_feel_smoke": value}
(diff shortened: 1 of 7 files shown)


Rasa_Bot/actions/actions_relapse_dialogs.py
```diff
@@ -897,11 +897,11 @@ def validate_how_feel_smoke(
         if last_utterance != 'utter_ask_how_feel_smoke':
             return {"how_feel_smoke": None}
 
-        valid = validator.validate_list(value, 0, 9)
+        valid = validator.validate_list(value, 1, 9)
 
         if not valid:
             dispatcher.utter_message(response="utter_did_not_understand")
-            dispatcher.utter_message(response="utter_please_answer_numbers")
+            dispatcher.utter_message(response="utter_please_answer_1_to_9")
```
nelealbers, line 904: I think the user can type multiple numbers between 1 to 9, so might be good to also reflect this in the message here rather than just asking them to type a number between 1 and 9. Something like "Type one or more numbers between 1 and 9 with a space in between"

Rasa_Bot/actions/actions_relapse_dialogs.py
```diff
@@ -1075,9 +1075,9 @@ def validate_hrs_feeling_slot(
         if last_utterance != 'utter_ask_hrs_feeling_slot':
             return {"hrs_feeling_slot": None}
 
-        is_valid = validator.validate_list(value, 1, 8)
+        is_valid = validator.validate_list(value, 1, 9)
         if not is_valid:
-            dispatcher.utter_message(response="utter_please_answer_1_to_8")
+            dispatcher.utter_message(response="utter_please_answer_1_to_9")
```
nelealbers, line 1080: Here also the user can type multiple values between 1 and 9 according to utter_ask_hrs_feeling_slot.

nelealbers (approved):
Looks good. For the two forms where users can type more than one number, I think it would be nice to also say that in the validation, and not just that they should type a single number between 1 and 9.

Rasa_Bot/actions/actions_relapse_dialogs.py
wbaccinelli: Good point. I change the two validations

PR #159, opened 2023-08-09
431 fixing weekly reflection pa

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/431

 - There was a bifurcation that made no sense, and it was not in the excel file.
 - Corrected a bug in the validation of a slot for the relapse dialog.
 - Corrected the default pa group.


nelealbers (approved):
Looks good. Took me some time to understand which branch is which, so could you also update the comment above the branch where you removed the group? Right now it still says "rule: show steps next week group1, 2". 

Also, in trying to understand the file on Teams, I saw that in the current implementation, the intensity group does not get lines 158 and 159? From the file on Teams it would seem to me though that both groups should get those two transition utterances. But maybe I misunderstand the file on Teams.

wbaccinelli:
> Looks good. Took me some time to understand which branch is which, so could you also update the comment above the branch where you removed the group? Right now it still says "rule: show steps next week group1, 2".
> 
> Also, in trying to understand the file on Teams, I saw that in the current implementation, the intensity group does not get lines 158 and 159? From the file on Teams it would seem to me though that both groups should get those two transition utterances. But maybe I misunderstand the file on Teams.

It is really complicated indeed. Both the groups get lines 158-159, but group 2 gets these utterances when exiting the HRS, so those are [in this dialog](https://github.com/PerfectFit-project/virtual-coach-rasa/blob/dcdf189af8382bd653a7fe006372a6c6d456ffc9/Rasa_Bot/data/rules/rules_relapse_dialog.yml#L928)

nelealbers:
> It is really complicated indeed. Both the groups get lines 158-159, but group 2 gets these utterances when exiting the HRS, so those are [in this dialog](https://github.com/PerfectFit-project/virtual-coach-rasa/blob/dcdf189af8382bd653a7fe006372a6c6d456ffc9/Rasa_Bot/data/rules/rules_relapse_dialog.yml#L928)

Ah I see, what a complicated dialog.

PR #161, opened 2023-08-10
404 tune intent recognition

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/404

- Increased threshold of fallback classifier.
- Removed confounding examples for the verder command.
- Added automated testing routine to verify the correct intent recognition.
- Verified with the misrecognized inputs reported in the testing tickets  that the intents are now correctly classified. 
- Removed unused stories for faster training.


Rasa_Bot/tests/test_conversations.yml
```diff
@@ -1,26 +1,61 @@
...
+
+- story: Test verder typo
+  steps:
+  - user: |
+      verderr
+    intent: continue_dialog_intent
+  - action: action_trigger_uncompleted_dialog
 
-- story: Ask foreseen HRS, user does not foresee HRS
+- story: Test fallback
```
nelealbers, line 54: Nice idea to add these tests. I remember that several of the wrong intent recognitions Eline and Anke encountered also came from sentences that contained one of the keywords (e.g., "Ik vind deze oefening heel stom"). I think it would be nice to also add a couple of these sentences to the fallback test.

Rasa_Bot/tests/test_conversations.yml
```diff
@@ -1,26 +1,61 @@
...
+  steps:
+  - user: |
+      verder
+    intent: continue_dialog_intent
+  - action: action_trigger_uncompleted_dialog
+
+- story: Test verder typo
+  steps:
+  - user: |
+      verderr
```
nelealbers, line 50: You removed "verderr" as an nlu example but are still using in the test stories?

nelealbers (approved):
Nice idea to add these test stories! Just see two comments below.

Rasa_Bot/tests/test_conversations.yml
wbaccinelli: That's a very good point, I will add a couple of those.

Rasa_Bot/tests/test_conversations.yml
wbaccinelli: Yes, removing the examples doesn't mean that the recognition is not general enough to recognize the typos. The problem was that with the set of examples used before, almost every word with few "r" was recognized as "verder".

nelealbers: Ah I see!

PR #164, opened 2023-08-11
add conditional response variations

nelealbers:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/413.
Fixes https://github.com/PerfectFit-project/testing-tickets/issues/56

Simple way of adding many conditions (looks ugly but unlikely to lead to errors).


wbaccinelli (approved):
Looks good. Having this very long list of conditions and alternative formulations probably makes the domain file a bit more difficult to read (not that was very readable before), but I guess that Rasa does not provide a more elegant way of writing this.

nelealbers:
> Looks good. Having this very long list of conditions and alternative formulations probably makes the domain file a bit more difficult to read (not that was very readable before), but I guess that Rasa does not provide a more elegant way of writing this.

The alternative is to create a new slot that just captures low and high values. But I did not want to add new logic.

PR #167, opened 2023-08-16
444 delay trigger fs

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/444

In order to avoid the messages of two consecutive dialogs to overlap in the execution phase: 

- Added a delay between the end of the weekly reflection and the beginning of the future self video.
- Added a delay between the end of the general activity dialog  and the beginning of the weekly reflection.


nelealbers (approved):
Okay!

PR #169, opened 2023-08-17
use description of activity before persuasive questions

nelealbers:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/426

- Use activity description before persuasive questions start so user has better idea of what activity is about.
- Also changed formulation according to part of this issue: https://github.com/PerfectFit-project/virtual-coach-issues/issues/407.


wbaccinelli (approved):
Well done!

PR #170, opened 2023-08-17
moved notification and fixed bug

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/427

- The notification for the quitting date has bene moved. It used to be planned after the relapse dialog, in case the quitting date changed. It is now sent on the quit date (or the new quit date) when reached. 
- Every time that the execution status was reached (after the prep phase or after the relapse) the execution week was reset and the general activity dialog was planned. This has been corrected to execute this behavior just if the execution status is reached after the preparation phase (execution week number 0). 
- On a side note, I removed the model training from the CI, as the tests were runnig on a model which was not the one actually used in the system. I also added a working model, trained with the latest rasa configuration.


nelealbers (approved):
So the buffer state is entered always after the goalsettingstate, which is called both at the start and in case of a relapse?

And what do you mean with "I also added a working model, trained with the latest rasa configuration."? You retrained the model?

wbaccinelli:
> So the buffer state is entered always after the goalsettingstate, which is called both at the start and in case of a relapse?

Correct, that's the state that goes from the setting of the quit date to the quit date itself. It's the same case in the preparation phase and after a relapse where a new quit date is set.
 
> And what do you mean with "I also added a working model, trained with the latest rasa configuration."? You retrained the model?
I retrained the model , otherwise the tests were failing.

PR #171, opened 2023-08-18
set timezone in controller

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/447

Modified the function in charge of creating the time values for the planning of dialogs in the controller. The times are now in the correct timezone.

### DIFF
--- a/scheduler/state_machine/state_machine_utils.py
+++ b/scheduler/state_machine/state_machine_utils.py
@@ -70,7 +70,8 @@ def create_new_date(start_date: date,
 
     """
     new_date = start_date + timedelta(days=time_delta)
-    new_timedate = datetime(new_date.year, new_date.month, new_date.day, hour, minute)
+    new_timedate = datetime(new_date.year, new_date.month, new_date.day, hour, minute,
+                            tzinfo=TIMEZONE)
 
     return new_timedate
 


nelealbers (approved):
Oh good catch! What about the other times that we use throughout the state machine? 

For example, we have comparisons like this one: 
` # (the long version has just been shown)
        if fs_time.date() <= date.today():`

`fs_time` now has the correct timezone, but `date.today()` would not be from the same timezone right? I do not know if those couple of hours difference matter here though - you certainly know that better.

And I noticed that there are statements such as `quit_datetime = create_new_date(quit_date).astimezone(TIMEZONE)`. The `astimezone` part would then no longer be needed?

wbaccinelli:
> Oh good catch! What about the other times that we use throughout the state machine?
> 
> For example, we have comparisons like this one: ` # (the long version has just been shown) if fs_time.date() <= date.today():`
> 
> `fs_time` now has the correct timezone, but `date.today()` would not be from the same timezone right? I do not know if those couple of hours difference matter here though - you certainly know that better.

The problem is there just when we are looking at the time, the dates should be fine anyway. Also, this is needed when interacting with the external word (so when planning the dialogs), if internal timings are uses, the localization is not needed.

wbaccinelli:
> @wbaccinelli Okay! And what about the addition I made to my comment about those astimezone things?

Ooops, I didn't read it. Your are right, that's redundant, good that you spotted this!

PR #174, opened 2023-08-31
458 split weekly reflection utterance

nelealbers:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/458
Fixes https://github.com/PerfectFit-project/testing-tickets/issues/78 (corresponding testing ticket)

- Split the utter_ask for the reflection form for PA so that the mooi/geweldig/oke part is not repeated when the form question is asked again, which happens when the user does not type enough words in their response.
- Did not retrain since small change.


wbaccinelli (approved):
Looks good!

PR #175, opened 2023-09-01
Add text about keeping intensity goal

bscheltinga:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/456
See: https://github.com/PerfectFit-project/testing-tickets/issues/62

### DIFF
--- a/Rasa_Bot/domain/domain_weekly_reflection_dialog.yml
+++ b/Rasa_Bot/domain/domain_weekly_reflection_dialog.yml
@@ -1064,13 +1064,13 @@ responses:
   utter_ask_if_doable:
     - text: "Zie jij dit zitten?\n
              Typ '1' voor ja.\n
-             Typ '2' voor nee. "
+             Typ '2' voor nee, ik behoud mijn intensiteitsdoel van vorige week. "
     - text: "Lijkt dit iets voor jou?\n
              Typ '1' voor ja.\n
-             Typ '2' voor nee. "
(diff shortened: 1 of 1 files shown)


Rasa_Bot/domain/domain_weekly_reflection_dialog.yml
```diff
@@ -1064,13 +1064,13 @@ responses:
   utter_ask_if_doable:
     - text: "Zie jij dit zitten?\n
              Typ '1' voor ja.\n
```
nelealbers, line 1066: Did Anke not also ask to change the 'ja' to 'ja, ik wil intensiever gaan bewegen'? Or was there a reason not to incorporate that? (you know about the goals better than I do)

nelealbers (approved):
Looks good, just see my comment below. I am not sure whether the 'ja' answer option also had to be adapted in the end or not.

Rasa_Bot/domain/domain_weekly_reflection_dialog.yml
bscheltinga: Yes, that was suggested. Though, when I went trough the dialog, if found this again to be confusing. Intensiever means: higher intensity, which is not really true in this case.

In the dialog, the new goal for intensity minutes is given, and the user is asked whether they like this or not. Then I think only 'ja' is sufficient. Also, when adding something like: "ja, ik weel meer intensief bewegen", users can intepret it as, even more than the given goal.

PR #178, opened 2023-09-05
452 missing step goal notification

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/452
Fixes https://github.com/PerfectFit-project/testing-tickets/issues/73

- Added the missing notifications between the quit date and the first general activity
- Default PA group to 1 in case no PA group has been set
- Grouped redundant code in a single function to schedule the PA notifications

### DIFF
--- a/scheduler/state_machine/controller.py
+++ b/scheduler/state_machine/controller.py
@@ -177,8 +177,11 @@ class OnboardingState(State):
         first_date = date.today() + timedelta(days=1)
         last_date = get_start_date(self.user_id) + timedelta(days=9)
 
-        plan_every_day_range(self.user_id,
-                             Notifications.TRACK_NOTIFICATION, 1, first_date, last_date)
+        plan_every_day_range(user_id=self.user_id,
+                             dialog=Notifications.TRACK_NOTIFICATION,
+                             phase_id=1,
+                             first_date=first_date,
+                             last_date=last_date)
 
 
 class TrackingState(State):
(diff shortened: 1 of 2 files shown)


scheduler/state_machine/controller.py
```diff
@@ -649,6 +627,18 @@ def schedule_pa_notifications(self):
                            planned_date=planned_date_4,
                            phase_id=2)
 
+        else:
+
+            first_date = date.today() + timedelta(days=1)
+            # until the next GA dialog
+            last_date = get_next_planned_date(self.user_id, datetime.now())
```
nelealbers, line 634: So get_next_planned_date() returns the next time the GA dialog is planned?

scheduler/state_machine/state_machine_utils.py
```diff
@@ -935,6 +935,31 @@ def plan_and_store(user_id: int,
...
+        phase_id: db id of the phase
+        first_date: first date where the dialog is planned
+        last_date: last date where the dialog is planned
+    """
+
+    for day in range((last_date - first_date).days + 1):
+        planned_date = create_new_date(start_date=first_date,
+                                       time_delta=day)
+
+        plan_and_store(user_id=user_id,
```
nelealbers, line 957: So why do we need plan_and_store for these notifications and could remove it from the quit date notification? Is that because the quit date notification is always sent immediately? 

And do we then check that these notifications are completed so we do not get the same issue that we had with the quit date?

wbaccinelli: Yes, exactly, in this case we need to store them to keep track of the planned tasks. In case a scheduled notification is sent, it is marked as completed in the celery task itself [here](https://github.com/PerfectFit-project/virtual-coach-rasa/blob/d57dd2035c76dd71f16fad4feb9d1d7f8a55297f/scheduler/celery_tasks.py#L336C9-L336C33)

PR #180, opened 2023-09-05
moved api url to env

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/461

- Moved the URLs definition to .env file
- Updated the prod URL


bscheltinga (approved):
Nice. I think we now should mention somewhere that the URLs should be in the env file.

wbaccinelli:
> Nice. I think we now should mention somewhere that the URLs should be in the env file.

@bscheltinga good suggestion, I updated the documentation

PR #181, opened 2023-09-06
Fix case no data available for step goal calculation

bscheltinga:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/460

What was going wrong: if there was no data from 9 days where the step goal should have been calculated, a NaN value was rounded, which gave the error. 

Fixed by checking if the data is empty or not, if empty, the goal is directly set to the minimum value.

### DIFF
--- a/sensor_api/sensorapi/connector.py
+++ b/sensor_api/sensorapi/connector.py
@@ -202,11 +202,15 @@ def get_step_goals_and_steps(steps_data: Optional[List[Dict[Any, Any]]],
         steps_nine_days = steps[i:i + 9]
         steps_nine_days = [x for x in steps_nine_days if not pd.isna(x)]  # Remove NaN values
 
-        for _ in range(9 - len(steps_nine_days)):
-            steps_nine_days.append(np.mean(steps_nine_days))  # Add the avg value up to 9 values
+        if not steps_nine_days:
+            step_goals.append(MIN_VALUE_STEP_GOAL)
 
-        steps_nine_days.sort()
-        step_goals.append(int(round(steps_nine_days[5], -1)))  # Definition of the goal
+        else:
+            for _ in range(9 - len(steps_nine_days)):
+                steps_nine_days.append(np.mean(steps_nine_days))  # Add the avg value up to 9 values
+
+            steps_nine_days.sort()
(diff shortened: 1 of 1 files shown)


sensor_api/sensorapi/connector.py
```diff
@@ -202,11 +202,15 @@ def get_step_goals_and_steps(steps_data: Optional[List[Dict[Any, Any]]],
         steps_nine_days = steps[i:i + 9]
         steps_nine_days = [x for x in steps_nine_days if not pd.isna(x)]  # Remove NaN values
 
-        for _ in range(9 - len(steps_nine_days)):
-            steps_nine_days.append(np.mean(steps_nine_days))  # Add the avg value up to 9 values
+        if not steps_nine_days:
+            step_goals.append(MIN_VALUE_STEP_GOAL)
```
wbaccinelli, line 206: in this case the length of `step_goals` is 1 instead of 9. Can this be a problem for further calculations/usage?

wbaccinelli (approved):
Nice catcht!

sensor_api/sensorapi/connector.py
bscheltinga: No, this is taken care of in:
 https://github.com/PerfectFit-project/virtual-coach-rasa/blob/f11a515650b5b00b18ee890a8fb6a005744974d6/sensor_api/sensorapi/connector.py#L204C1-L207 

Only the case of no data at all during 9 consecutive days was not covered.

wbaccinelli: I see, but that line is not  executed if `steps_nine_days` was empty, so the returned `step_goals` value is an int instead of a list. But maybe I am mistaking something, or it is not important.

bscheltinga: I see your point, but how `steps_nine_days`  is defined earlier in the code, it is always a list as far I understand:
https://github.com/PerfectFit-project/virtual-coach-rasa/blob/f11a515650b5b00b18ee890a8fb6a005744974d6/sensor_api/sensorapi/connector.py#L199-L203

I think even if all values but 1 are NaN, the output will be a list with a single value.

wbaccinelli: Ok, I think I got the point. Thanks!

PR #183, opened 2023-09-06
462 get smoking status

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/462

- get tracked cigarettes from NiceDay
- get reported cigarettes from (re)lapses
- check smoking status in closing dialog

### DIFF
--- a/Rasa_Bot/actions/helper.py
+++ b/Rasa_Bot/actions/helper.py
@@ -1347,17 +1347,20 @@ def get_smoked_cigarettes_range(user_id: int,
 
     cigarettes = 0
 
-    cigarettes_entries = (session.query(DialogOpenAnswers)
-                  .join(DialogQuestions)
-                  .filter(DialogOpenAnswers.users_nicedayuid == user_id,
-                          ((
-                                   DialogQuestions.question_id == DialogQuestionsEnum.RELAPSE_LAPSE_NUMBER_CIGARETTES.value) | (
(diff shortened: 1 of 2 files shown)


Rasa_Bot/actions/helper.py
```diff
@@ -1328,6 +1329,46 @@ def get_steps_data(user_id: int,
...
+        session.query(DialogOpenAnswers)
+        .join(DialogQuestions)
+        .filter(DialogOpenAnswers.users_nicedayuid == user_id,
+                ((
+                    DialogQuestions.question_id == DialogQuestionsEnum
+                    .RELAPSE_LAPSE_NUMBER_CIGARETTES.value) | (
+                    DialogQuestions.question_id == DialogQuestionsEnum
+                    .RELAPSE_RELAPSE_NUMBER_CIGARETTES.value)),
+                DialogOpenAnswers.datetime >= start_date,
+                DialogOpenAnswers.datetime <= end_date
```
nelealbers, line 1360: So the end_date is also included in the interval? I am asking because the documentation says that end_date is not included in the interval.

nelealbers (approved):
Looks good! Just one minor question below. Also, you removed the old model but added no new one?

Rasa_Bot/actions/helper.py
wbaccinelli: Yes, that was a mistake in the docstring, thanks for spotting it!

wbaccinelli:
> Looks good! Just one minor question below. Also, you removed the old model but added no new one?

I didn't realized that I removed the model.. I restored it now

PR #187, opened 2023-09-14
Change step algorithm, use 0 instead of average

bscheltinga:
Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/466

### DIFF
--- a/sensor_api/sensorapi/connector.py
+++ b/sensor_api/sensorapi/connector.py
@@ -192,30 +192,26 @@ def get_step_goals_and_steps(steps_data: Optional[List[Dict[Any, Any]]],
     if df.index.max().strftime('%y%m%d') != end_date.strftime('%y%m%d'):
         df.loc[end_date.date()] = np.nan
 
-    # Resample the data to a day-to-day basis. Add nans for empty dates
-    df = df.asfreq('D')
+    # Resample the data to a day-to-day basis. Add zeros for empty dates
+    df = df.asfreq('D', fill_value=0)
 
     # Calculate the step goals
     steps = list(df.steps)
     step_goals = []
     for i in range(7):
         steps_nine_days = steps[i:i + 9]
-        steps_nine_days = [x for x in steps_nine_days if not pd.isna(x)]  # Remove NaN values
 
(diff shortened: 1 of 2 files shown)


sensor_api/sensorapi/connector.py
```diff
@@ -192,30 +192,26 @@ def get_step_goals_and_steps(steps_data: Optional[List[Dict[Any, Any]]],
...
-    df = df.asfreq('D')
+    # Resample the data to a day-to-day basis. Add zeros for empty dates
+    df = df.asfreq('D', fill_value=0)
 
     # Calculate the step goals
     steps = list(df.steps)
     step_goals = []
     for i in range(7):
         steps_nine_days = steps[i:i + 9]
-        steps_nine_days = [x for x in steps_nine_days if not pd.isna(x)]  # Remove NaN values
```
wbaccinelli, line 203: I think that we can keep this as an extra security layer, so that if for any reason we have NaNs we get rid of them.

sensor_api/sensorapi/connector.py
```diff
@@ -192,30 +192,26 @@ def get_step_goals_and_steps(steps_data: Optional[List[Dict[Any, Any]]],
...
 
         if not steps_nine_days:
             step_goals.append(MIN_VALUE_STEP_GOAL)
 
         else:
-            for _ in range(9 - len(steps_nine_days)):
-                steps_nine_days.append(np.mean(steps_nine_days))  # Add the avg value up to 9 values
-
```
wbaccinelli, line 211: If we decide to keep the NaN values check we should keep this as well, right?

wbaccinelli (approved):
Good catch! I just left few comments, if we want to keep a verification on the NaN values as an extra safety layer.

sensor_api/sensorapi/connector.py
bscheltinga: I added it again. But I think there are no NaN values anymore.

sensor_api/sensorapi/connector.py
bscheltinga: Yes, I change it :)

PR #191, opened 2023-09-21
470 fak video non mandatory

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/470
Fixes https://github.com/PerfectFit-project/testing-tickets/issues/87

- The fsm goes directly to the execution state in case the quit date has been reached, even if the first aid kit video has not been completed
- Save the status when running it, to make sure that the correct state has been stored

### DIFF
--- a/scheduler/state_machine/controller.py
+++ b/scheduler/state_machine/controller.py
@@ -291,6 +291,9 @@ class GoalsSettingState(State):
                           planned_date=new_date,
                           phase=1)
 
+    def on_new_day(self, current_date: date):
+        self.check_if_end_date(current_date)
+
     def on_user_trigger(self, dialog):
         # the relapse dialog is not available in this phase
         if dialog == Components.CONTINUE_UNCOMPLETED_DIALOG:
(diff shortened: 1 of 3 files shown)


scheduler/state_machine/state_machine_utils.py
```diff
@@ -765,6 +765,28 @@ def retrieve_tracking_day(user_id: int, current_date: date) -> int:
...
+                .filter(UserStateMachine.users_nicedayuid == user_id)
+                .first())
+
+    # if a machine already exists, use the same primary key to update the row
+    if user_fsm is not None:
+        user_fsm.state = state
+
+        session.merge(user_fsm)
+        session.commit()
+
```
bscheltinga, line 788: Don't you need the `session.close()` here?

scheduler/state_machine/controller.py
```diff
@@ -291,6 +294,9 @@ def on_dialog_rescheduled(self, dialog, new_date):
                           planned_date=new_date,
                           phase=1)
 
+    def on_new_day(self, current_date: date):
+        self.check_if_end_date(current_date)
+
```
bscheltinga, line 299: When it this called?

bscheltinga (approved):

scheduler/state_machine/state_machine_utils.py
wbaccinelli: I absolutely do, good catch!

scheduler/state_machine/controller.py
wbaccinelli: This is triggered when a `new_day` event is sent by the FSM. This happens every morning.

bscheltinga: Check, I see it now in state_machine.py :)

scheduler/state_machine/state_machine_utils.py
bscheltinga: 🙏

PR #195, opened 2023-10-04
check completion state of 29 activity

wbaccinelli:
Fix https://github.com/PerfectFit-project/virtual-coach-issues/issues/474

### DIFF
--- a/scheduler/state_machine/controller.py
+++ b/scheduler/state_machine/controller.py
@@ -241,7 +241,8 @@ class TrackingState(State):
         # at day 7 activity C2.9 has to be proposed
         start_date = get_start_date(self.user_id)
         choose_sport_completed = get_activity_completion_state(self.user_id, 29)
-        if (current_date - start_date).days >= ACTIVITY_C2_9_DAY_TRIGGER and not choose_sport_completed:
+        if ((current_date - start_date).days >= ACTIVITY_C2_9_DAY_TRIGGER
+                and not choose_sport_completed):
             plan_and_store(user_id=self.user_id,
                            dialog=Components.GENERAL_ACTIVITY,
                            phase_id=1)
--- a/scheduler/state_machine/controller.py
+++ b/scheduler/state_machine/controller.py
@@ -240,8 +240,8 @@ class TrackingState(State):
 
         # at day 7 activity C2.9 has to be proposed
         start_date = get_start_date(self.user_id)
-        ga_completed = get_activity_completion_state(self.user_id, 29)
-        if (current_date - start_date).days >= ACTIVITY_C2_9_DAY_TRIGGER and not ga_completed:
+        choose_sport_completed = get_activity_completion_state(self.user_id, 29)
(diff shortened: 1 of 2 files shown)


scheduler/state_machine/controller.py
```diff
@@ -239,7 +240,7 @@ def on_new_day(self, current_date: date):
 
         # at day 7 activity C2.9 has to be proposed
         start_date = get_start_date(self.user_id)
-        ga_completed = get_dialog_completion_state(self.user_id, Components.GENERAL_ACTIVITY)
+        ga_completed = get_activity_completion_state(self.user_id, 29)
         if (current_date - start_date).days >= ACTIVITY_C2_9_DAY_TRIGGER and not ga_completed:
```
bscheltinga, line 244: Can't we use a enumerator here like it was with `Components.GENERAL_ACTIVITY`

scheduler/state_machine/state_machine_utils.py
```diff
@@ -76,6 +77,40 @@ def create_new_date(start_date: date,
...
+        .limit(1)  # get only the first result
+        .one_or_none()
+    )
+
+    if activities is not None:
+        session.close()
+        return True
+
+    session.close()
+    return False
```
bscheltinga, line 111: I think you can close the session before the if-loop?

scheduler/state_machine/controller.py
```diff
@@ -239,7 +240,7 @@ def on_new_day(self, current_date: date):
 
         # at day 7 activity C2.9 has to be proposed
         start_date = get_start_date(self.user_id)
-        ga_completed = get_dialog_completion_state(self.user_id, Components.GENERAL_ACTIVITY)
+        ga_completed = get_activity_completion_state(self.user_id, 29)
```
bscheltinga, line 243: Change variable name `ga_completed` into something more suitable like `choose_sport_completed`

scheduler/state_machine/controller.py
```diff
@@ -332,7 +333,7 @@ def plan_execution_start_dialog(self):
         # this is the intro video to be sent the first time
         # the execution starts (not after lapse/relapse)
         quit_date = get_quit_date(self.user_id)
-        planned_date = create_new_date(start_date=quit_date)
+        planned_date = create_new_date(start_date=quit_date, minute=5)
```
bscheltinga, line 336: Not sure why this is added?

bscheltinga (approved):
Hi Walter, I have some questions/remarks for you!

scheduler/state_machine/state_machine_utils.py
wbaccinelli: No, because if the session is closed, the `activities` object is emptied, as it's linked to the session.

scheduler/state_machine/controller.py
wbaccinelli: I realized that this dialog uses the same time of the notifications, and the two things would overlap

PR #201, opened 2023-11-15
fix: schedule weekly reflection when not automatically scheduled

wbaccinelli:
Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/482

During the execution phase, every day the the state machine checks if the weekly reflection dialog has been planned for the current week. It retrieves the last date when the weekly reflection should have been sent and, if no new weekly reflection has been sent nor planned, it triggers the dialog.

### DIFF
--- a/scheduler/state_machine/state_machine_utils.py
+++ b/scheduler/state_machine/state_machine_utils.py
@@ -57,6 +57,31 @@ def compute_next_day(selectable_days: list, current_date: datetime) -> date:
     return next_date
 
 
+def compute_previous_day(user_id: int, current_date: datetime) -> date:
+    """
+    Gets the date of the last preferred day for a user.
+
+    Args:
+        user_id: the id of the user
+        current_date: the date to start from
+
+    Returns:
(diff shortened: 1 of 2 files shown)


scheduler/state_machine/state_machine_utils.py
```diff
@@ -57,6 +57,31 @@ def compute_next_day(selectable_days: list, current_date: datetime) -> date:
...
+        current_date: the date to start from
+
+    Returns:
+        The date of the last preferred day for a user.
+    """
+
+    date_time = get_preferred_date_time(user_id=user_id)
+
+    # first element of the tuple is the list of days
+    pref_day = date_time[0][0]
```
BendikC, line 75: Looks like this could cause an error if left unhandled, saw that get_preferred_date_time could potentially return (None, None) so this would fail in this case.

wbaccinelli: this should not happen, since the value should have been set in the profile creation. But it is indeed good to add a control here. I will do it.

PR #202, opened 2023-11-20
General activity restart dialog

BendikC:
Two changes made:
- Verder implementation changed to prefer other dialogues over general activity
- New day changed to run uncompleted general activity rather than storing a new one

### DIFF
--- a/scheduler/state_machine/controller.py
+++ b/scheduler/state_machine/controller.py
@@ -17,7 +17,8 @@ from state_machine.state_machine_utils import (create_new_date, get_dialog_compl
                                                save_fsm_state_in_db,
                                                schedule_next_execution, store_completed_dialog,
                                                store_scheduled_dialog, update_execution_week,
-                                               update_fsm_dialog_running_status)
+                                               update_fsm_dialog_running_status,
+                                               dialogs_to_be_completed, get_component_id)
 from state_machine.const import (ACTIVITY_C2_9_DAY_TRIGGER, FUTURE_SELF_INTRO, GOAL_SETTING,
                                  TRACKING_DURATION, TIMEZONE, PREPARATION_GA, PAUSE_AND_TRIGGER,
                                  MAX_PREPARATION_DURATION, HIGH_PA_GROUP,
@@ -243,9 +244,12 @@ class TrackingState(State):
         choose_sport_completed = get_activity_completion_state(self.user_id, 29)
         if ((current_date - start_date).days >= ACTIVITY_C2_9_DAY_TRIGGER
                 and not choose_sport_completed):
-            plan_and_store(user_id=self.user_id,
-                           dialog=Components.GENERAL_ACTIVITY,
-                           phase_id=1)
+            if self.check_if_general_activity_dialog_exists():
+                run_uncompleted_dialog(self.user_id, dialog_preference=Components.GENERAL_ACTIVITY)
+            else:
+                plan_and_store(user_id=self.user_id,
+                               dialog=Components.GENERAL_ACTIVITY,
+                               phase_id=1)
 
         self.check_if_end_date(current_date)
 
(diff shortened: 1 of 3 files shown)


scheduler/state_machine/controller.py
```diff
@@ -258,6 +262,16 @@ def check_if_end_date(self, date_to_check: date) -> bool:
...
 
+    def check_if_general_activity_dialog_exists(self):
+        """
+        Check if there is an uncompleted general activity dialog in the database for the given user.
+        """
+        general_activity_id = get_component_id(Components.GENERAL_ACTIVITY)
+        uncompleted_dialogs = dialogs_to_be_completed(self.user_id)
+        dialog_ids = list(map(lambda dialog: dialog.intervention_component_id, uncompleted_dialogs))
+
+        return general_activity_id in dialog_ids
```
wbaccinelli, line 273: this is used just to check if the there is an uncompleted dialog, right? Maybe it is not needed to return the ids, but just a bool.

scheduler/state_machine/state_machine_utils.py
```diff
@@ -778,29 +778,28 @@ def dialog_to_be_completed(user_id: int) -> Optional[UserInterventionState]:
             UserInterventionState.completed.is_(False),
             UserInterventionState.last_time.isnot(None)
         )
-        .first()
+        .all()
     )
 
-    session.expunge(uncompleted)
+    for dialog in uncompleted:
+        session.expunge(dialog)
```
wbaccinelli, line 785: This is tricky, because with lists, expunge doesn't work properly. You have to remap the values like:         

```suggestion
    result = [UserInterventionState(id=task.id,
                                        users_nicedayuid=task.users_nicedayuid,
                                        intervention_phase_id=task.intervention_phase_id,
                                        intervention_component_id=task.intervention_component_id,
                                        completed=task.completed,
                                        last_time=task.last_time,
                                        last_part=task.last_part,
                                        next_planned_date=task.next_planned_date,
                                        task_uuid=task.task_uuid) for task in uncompleted]
```

scheduler/state_machine/state_machine_utils.py
```diff
@@ -778,29 +778,28 @@ def dialog_to_be_completed(user_id: int) -> Optional[UserInterventionState]:
...
 
     Args:
         user_id: ID of the user
+        dialog_preference: Dialog to prioritize when selecting which one to run
 
     """
-
-    uncompleted = dialog_to_be_completed(user_id)
+    uncompleted = dialogs_to_be_completed(user_id)
+    uncompleted = select_dialog_to_complete(uncompleted, dialog_preference)
```
wbaccinelli, line 802: if no dialogs have to be completed, are you sure that you obtain an empty list and not a None value?

scheduler/state_machine/state_machine_utils.py
```diff
@@ -821,6 +820,60 @@ def run_uncompleted_dialog(user_id: int):
...
+        .first()
+    )
+
+    session.close()
+
+    if intervention_component is None:
+        logging.error(dialog.value, " not found in the database")
+        return None
+
+    return intervention_component.intervention_component_id
```
wbaccinelli, line 875: When the session is closed you have no more access to the query result, so you have to store the value you want to return in a new variable.

```suggestion
    id = intervention_component.intervention_component_id
    session.close()

    if intervention_component is None:
        logging.error(dialog.value, " not found in the database")
        return None

    return id

```

wbaccinelli (approved):
Well done! I left a few comments, they are mostly related to the management of db sessions, which are not testable locally, because we use the managed db only in prod.

scheduler/state_machine/controller.py
BendikC: Yes, here we return a bool and not an ID, the "in" keyword makes it a bool by seeing if the if the id is in the list and returning true if it is or false otherwise.

scheduler/state_machine/state_machine_utils.py
BendikC: It's not too clear in the docs even though it does seem to be the case. Chatgpt also agrees but it still could cause issues so I will handle the case in select_dialog_to_complete

scheduler/state_machine/controller.py
wbaccinelli: Oh, good. You can declare the returned data type for clarity, if you want

PR #204, opened 2023-12-14
Adds the notification to tracking state

BendikC:
From what I am seeing, the tracking state notification is a daily notification triggered after the completion of the tracking dialog, and it's the only place after the onboarding where we don't send notifications to the user to ensure that they wear the watch and collect data. 
It seems as if adding this single line should do the trick, from what I've looked at, but I'm not completely sure, so I also wouldn't mind making a different notification to add to this state.

### DIFF
--- a/Rasa_Bot/data/rules/rules_notifications.yml
+++ b/Rasa_Bot/data/rules/rules_notifications.yml
@@ -8,6 +8,7 @@ rules:
   - action: utter_notifications_tracking_1
   - action: utter_notifications_tracking_2
   - action: utter_notifications_tracking_3
+  - action: action_notifications_check_watch_wear
   - action: pause_one_minute
 
 ## Step goal notification


bscheltinga:
@wbaccinelli do you know the tracking notification is triggered daily after the tracking dialog?

bscheltinga (approved):
Nice that it is fixed this easily :)

bscheltinga:
@wbaccinelli @BendikC I will discuss the content of the notifications itself with Milon and Roxy as well. I think it can be phrased in a more active form, hopefully it will motivate people more to do it then.

bscheltinga:
@BendikC can you merge this? We should also notify Walter that he should add this to the production server

wbaccinelli:
https://github.com/PerfectFit-project/virtual-coach-issues/issues/485

PR #205, opened 2023-12-14
Mark dialogs as completed after ending dialog

BendikC:
User got notifications after a dialog was ended for this particular dialog, which implied that the dialog hadn't been marked as completed. 'action_end_dialog' is deceptive as it doesn't really do anything right now, but mark_dialog_completed is important to conclude with so the State Machine updates the dialog as not running anymore.


wbaccinelli (commented):
Good catch! I just wonder weather 2 followup actions can be returned. Have you tried it or is there any other equivalent situation?

BendikC:
Actually a good point, I assumed since it was a list and you can do multiple slotsets that actions would be possible too, but I'm reading now that this might not be the case.

PR #206, opened 2024-01-08
fix: empty graphs fix

BendikC:
Fix https://github.com/PerfectFit-project/virtual-coach-issues/issues/486
The issue was that we were not copying over the closed_answers_id so it was always none


wbaccinelli (approved):
Very well done!

wbaccinelli:
Fix https://github.com/PerfectFit-project/virtual-coach-issues/issues/486

PR #207, opened 2024-01-08
fix: check execution week 12 instead of 11

BendikC:
Fix https://github.com/PerfectFit-project/virtual-coach-issues/issues/488
changed the check


wbaccinelli (approved):
Thanks for fixing this!

PR #208, opened 2024-01-09
hotfix: new model

wbaccinelli:
Fix https://github.com/PerfectFit-project/virtual-coach-issues/issues/489


BendikC (approved):
Good catch, forgot that we also needed to retrain after adding the rule

PR #209, opened 2024-01-09
fix: set slot before week number check in weekly reflection

wbaccinelli:
Fix https://github.com/PerfectFit-project/virtual-coach-issues/issues/490


BendikC (approved):
Good catch! This way we don't change the intervention component to weekly reflection before potentially leaving the dialog on the next action

PR #211, opened 2024-01-16
return to weekly reflection successfully

BendikC:
Added the followupaction so we return to the right place, and added two rules so that rasa doesn't get confused.

Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/493


wbaccinelli (approved):
Looks good, well done!

PR #212, opened 2024-01-16
Check if user changed quit date

BendikC:
If user changed quit date we trigger the same logic that we would trigger from the relapse dialog, but after the completion of the weekly reflection dialog.

Fixes: https://github.com/PerfectFit-project/virtual-coach-issues/issues/492


scheduler/state_machine/controller.py
```diff
@@ -595,6 +595,23 @@ def on_dialog_completed(self, dialog):
...
+                # if a new quit date has been set, the weekly reflection might be rescheduled,
+                # a notification on the day before and on the new date are planned.
+                # Then we go back to the buffer state
+                self.reschedule_weekly_reflection(quit_date)
+                self.plan_new_date_notifications(quit_date)
+                self.set_new_state(BufferState(self.user_id))
+                store_completed_dialog(user_id=self.user_id,
+                                       dialog=Components.RELAPSE_DIALOG,
+                                       phase_id=3)
+
```
wbaccinelli, line 614: This is good, but I would move it to line 530, and avoid planning the new dialogs in this case.
Also, the `reschedule_weekly_reflection` and `plan_new_date_notifications` are defined only for the `RelapseState` class, so they should either moved to the `state_machine_utils` and imported, or redefined for this class (no a clean solution)

wbaccinelli (changes requested):
The solution is correct, but I left a comment for a couple of needed modifications. It should also solve the CI issue.

wbaccinelli (approved):
Thanks for the changes, looks good.

PR #213, opened 2024-01-16
add if check

BendikC:
only save to db if we are in goal setting dialog.

Fixes https://github.com/PerfectFit-project/virtual-coach-issues/issues/491


wbaccinelli (approved):
Looks good!

